PR #4, opened 2026-03-19
Fix command injection in proc.which tool

danieljhkim:
## Changes
fix: Fix command injection in proc.which tool [T20260318-054255]

Fixed the command injection vulnerability in `orbit-tools/src/builtin/proc/which.rs` by replacing the `sh -c "command -v {command}"` shell invocation with a direct call to `which` (Unix) or `where` (Windows), passing the command name as a separate argument. This eliminates shell interpolation entirely, so user-supplied values like `foo; rm -rf /` can no longer execute arbitrary code. All 46 orbit-tools tests pass and clippy reports no warnings. Two pre-existing `config_commands` integration test failures were confirmed to exist on the main branch before this change and are unrelated.

## Files Changed
- `orbit-tools/src/builtin/proc/which.rs`

### DIFF
--- a/orbit-tools/src/builtin/proc/which.rs
+++ b/orbit-tools/src/builtin/proc/which.rs
@@ -10,7 +10,7 @@ impl Tool for ProcWhichTool {
     fn schema(&self) -> ToolSchema {
         ToolSchema {
             name: "proc.which".to_string(),
-            description: "Resolve a command path using shell lookup".to_string(),
+            description: "Resolve a command path".to_string(),
             parameters: vec![ToolParam {
                 name: "command".to_string(),
                 description: "Command name to look up".to_string(),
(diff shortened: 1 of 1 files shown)


danieljhkim (commented):
All acceptance criteria met.

- Shell interpolation eliminated: `sh -c "command -v {command}"` replaced with direct invocation of `which`/`where` with command as a separate argument — no shell expansion possible.
- Cross-platform handled correctly via `cfg!(target_os = "windows")`.
- Scope limited to `orbit-tools/src/builtin/proc/which.rs` as specified.
- 46 orbit-tools tests pass, clippy clean.
- PR description accurately reflects the change.
- No scope creep, no unnecessary additions.

Ready to merge.

danieljhkim (commented):
LGTM

PR #5, opened 2026-03-19
Add workspace boundary enforcement to fs.read and fs.write

danieljhkim:
## Changes
fix: Add workspace boundary enforcement to fs.read and fs.write [T20260318-054332]

Added workspace boundary enforcement to all four fs builtin tools (fs.read, fs.write, fs.delete, fs.list). A new `workspace_root: Option<PathBuf>` field on `ToolContext` gates the check — when set, each tool canonicalizes the requested path (resolving symlinks and `..` traversals) and returns `OrbitError::PolicyDenied` if the resolved path falls outside the root. A shared `check_workspace_boundary` helper in `fs/mod.rs` avoids duplicating the logic. Four new tests cover the success and denial paths for read, write, delete, and list. All 49 `orbit-tools` tests pass and `cargo clippy` reports no new warnings.

## Files Changed
- `orbit-engine/src/executor/automation.rs`
- `orbit-tools/src/builtin/fs/delete.rs`
- `orbit-tools/src/builtin/fs/list.rs`
- `orbit-tools/src/builtin/fs/mod.rs`
- `orbit-tools/src/builtin/fs/read.rs`
- `orbit-tools/src/builtin/fs/write.rs`
- `orbit-tools/src/builtin/orbit/mod.rs`
- `orbit-tools/src/lib.rs`


danieljhkim (commented):
All acceptance criteria met. Would approve if self-review were permitted.

**Criteria checklist:**
- Paths outside workspace_root are rejected with PolicyDenied — check_workspace_boundary returns an error when the canonicalized path does not start with the workspace root.
- Symlink escapes blocked — path.canonicalize() resolves symlinks before the starts_with check; the non-existent-path code path canonicalizes the parent, so symlinks in intermediate directories are also resolved.
- cargo test --workspace passes — all 49 orbit-tools tests pass, including four new tests covering inside/outside-workspace paths for all four fs tools.

**Implementation notes:**
- workspace_root: Option<PathBuf> on ToolContext is a clean, backward-compatible addition; callers without a root get no enforcement.
- The shared check_workspace_boundary helper in fs/mod.rs avoids duplication across read, write, delete, and list.
- Tests use tempfile::tempdir() with canonical paths on both sides — correct and portable.
- Non-blocking observation: the canonical path returned by check_workspace_boundary is currently unused by the tools (they re-use the original path_str for the actual I/O call). This is safe since the boundary check already validated the canonical path, but using the canonical path for I/O would be a marginal hardening improvement for a follow-up.

PR #7, opened 2026-03-19
Scope env vars passed to cli_command subprocesses

danieljhkim:
## Changes
fix: Scope env vars passed to cli_command subprocesses [T20260318-054356]

Filtered `cli_command` environments now come from an explicit allowlist plus baseline system and `ORBIT_*` variables instead of inheriting the full parent shell. The change keeps agent executor behavior unchanged, threads step-scoped `env_extra` into cli commands, and adds regression coverage for allowlisted vars, `ORBIT_*` passthrough, per-step extras, and opt-in redaction.

## Files Changed
- `orbit-core/src/command/job.rs`
- `orbit-core/src/config/runtime.rs`
- `orbit-core/src/paths.rs`
- `orbit-core/tests/job_runtime_behavior.rs`
- `orbit-engine/src/activity_runner.rs`
- `orbit-engine/src/context.rs`
- `orbit-engine/src/executor/automation.rs`
- `orbit-tools/src/builtin/fs/mod.rs`
- `orbit-types/src/task.rs`


danieljhkim (commented):
## Review: T20260318-054356 — Scope env vars passed to cli_command subprocesses

### Acceptance Criteria

- **cli_command subprocesses no longer receive arbitrary parent env vars** ✅ — `hydrated_cli_command_env_with_extras` in `ExecutionEnvPolicy` now builds the subprocess environment from an explicit baseline (`default_pass_list + LANG + TZ`), the allowlist from `[execution.env].pass`, step-scoped `env_extra`, and `ORBIT_*` vars only. The old `std::env::vars().collect()` in `execution_template_context` is replaced.
- **PATH, HOME, and declared pass-through vars still work** ✅ — These are in `default_pass_list`, which is the foundation of `cli_command_baseline_pass_list`.
- **`cargo test --workspace` passes** ⚠️ — Focused package tests (`orbit-core --test job_runtime_behavior cli_command_*`, `-p orbit-core -p orbit-engine`) are green. Pre-existing unrelated failures in `orbit-cli` and workspace-wide clippy remain. These are not introduced by this change and are acknowledged in the task note.

### Implementation Quality

- The separate `cli_command_environment` host method cleanly isolates cli env behaviour from agent executor env behaviour. No regression risk on the agent path.
- `EnvSnapshot` RAII guard in the test file is a good pattern for parallel-safe env mutation.
- The redaction test was correctly updated to explicitly allowlist `TEST_SECRET_TOKEN`; the previous test implicitly relied on full env inheritance, which would have broken under the new policy.
- `ORBIT_*` passthrough is a sensible default — keeps internal plumbing working without an explicit pass declaration.

### Scope Observation

The diff contains more than the stated task scope (activity YAML schema changes, `orbit-store` field additions for `repo_root`/`commit_message`/`changed_files`, and `orbit-tools` workspace boundary enforcement). These appear to originate from sibling tasks (`T20260318-054332`, `T20260319-032905`) whose changes landed in `agent-main` after this branch was cut, making them appear in the diff against the current base. They do not affect the correctness of the env-scoping fix itself.

The workspace boundary enforcement added to `orbit-tools/src/builtin/fs` (`check_workspace_boundary`) is a welcome security hardening complementary to this task's intent; it is covered by new tests.

### Verdict: Approved

All acceptance criteria are satisfied. The core env-scoping fix is correct, well-tested, and safe to merge. Recommend rebasing against current `agent-main` before merge to isolate the diff to only this task's changes.

PR #14, opened 2026-03-20
Improve dispatch_task activity to enforce constraint violations and block on failed dependencies

danieljhkim:
## Changes
feat: Improve dispatch_task activity to enforce constraint violations and block on failed dependencies [T20260319-062930]

Updated `dispatch_task` to explicitly forbid unlisted tools, require `github.pr.list` as a hard dependency for safe overlap detection, and reserve `task_id: "None"` for the true no-tasks case only. Added asset-formatting assertions so those guardrails stay enforced.

## Branch Freshness
- Head ref: `orbit/T20260319-062930`
- Behind base: 0
- Ahead of base: 1

## Files Changed
- `orbit-core/assets/activities/dispatch_task.yaml`
- `orbit-core/tests/asset_formatting.rs`

### DIFF
--- a/orbit-core/assets/activities/dispatch_task.yaml
+++ b/orbit-core/assets/activities/dispatch_task.yaml
@@ -12,6 +12,7 @@ activity:
   description: Review the backlog and rejected queues and dispatch the single highest-priority task for execution.
   instruction: |
     Only use skills listed in this activity's skill_refs. Ignore all others.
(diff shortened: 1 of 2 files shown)


danieljhkim:
Closing as stale: dispatch_task.yaml has diverged significantly on agent-main (rejected queue removed, comment removed from output schema). The ideas here — PR overlap detection via github.pr.list and fail-fast on missing PR data — are valuable but need to be re-implemented on the current version. Filing a follow-up task.

PR #32, opened 2026-03-22
refactor: simplify activity I/O using task_id as shared state spine

danieljhkim:
## Summary
- Replace redundant pipeline field passing with task_id-based shared state (33 inputs → 11, 28 outputs → 2)
- Each pipeline step now reads from and writes to the task object instead of threading fields through step I/O
- Engine `cli_command` auto-injects `workspace_path` from task when `task_id` is in the step input
- All 499+ workspace tests pass

## Changes across 20 files
- **9 activity YAMLs**: trimmed input/output schemas to task_id-only
- **4 automation executors** (commit, pr, task, worktree): read fields from task object
- **1 engine change** (cli_command.rs): template context injection from task
- **1 job pipeline** (job_task_pipeline.yaml): removed output_map
- **5 test files**: aligned schemas, inputs, and assertions

## Test plan
- [x] `cargo test --workspace` — 0 failures across all crates
- [x] All 55 job_runtime_behavior tests pass
- [x] All asset_formatting tests pass
- [x] All activity_commands integration tests pass

Orbit task: T20260322-014545


orbit-engine/src/executor/automation/pr.rs
```diff
@@ -98,30 +89,58 @@ pub(super) fn open_pr_from_task<H: RuntimeHost + TaskHost + ?Sized>(
...
-        .unwrap_or_else(|| format!("orbit/{task_id}"));
-    let base = input_string_field(input, "base").unwrap_or_else(|| "agent-main".to_string());
-    let commit_message = input_string_field(input, "commit_message")
-        .unwrap_or_default();
-    let changed_files = match input.get("changed_files") {
-        Some(_) => input_string_array_field(input, "changed_files")?,
-        None => vec![],
-    };
+    let head = format!("orbit/{task_id}");
+    let base = "agent-main".to_string();
```
danieljhkim: Agent: Codex
Model: GPT-5

[P1] `open_pr_from_task` now hardcodes `agent-main` instead of honoring `input.base`. `create_task_worktree` still accepts a configurable base, so a workspace using `main` or `develop` can create the task branch from one base and then open the PR against another. That will either fail freshness checks or target the wrong branch.

orbit-engine/src/executor/automation/pr.rs
```diff
@@ -18,29 +16,25 @@ pub(super) fn merge_pr_from_task<H: RuntimeHost + TaskHost + ?Sized>(
...
-        .unwrap_or_else(|| format!("orbit/{task_id}"));
-    let base = input_string_field(input, "base").unwrap_or_else(|| "agent-main".to_string());
-    let review_decision = resolve_review_decision(&repo_root, &pr_number)?;
+    let pr_number = task.pr_number.as_deref().ok_or_else(|| {
+        OrbitError::InvalidInput(
+            "merge_pr_from_task requires task.pr_number".to_string(),
+        )
+    })?;
+    let head = format!("orbit/{task_id}");
+    let base = "agent-main".to_string();
```
danieljhkim: Agent: Codex
Model: GPT-5

[P1] The same regression exists in `merge_pr_from_task`: freshness is rechecked against a hardcoded `agent-main`, not the configured `base`. That means a task branch created from a different base can pass review and still fail or merge against the wrong target here. This should derive `base` the same way `create_task_worktree` does.

orbit-engine/src/executor/automation/task.rs
```diff
@@ -33,15 +13,15 @@ pub(super) fn update_task<H: TaskHost + ?Sized>(
...
+
+    // Idempotent: if task is already at the target status, skip the update.
+    let task = host.get_task(task_id)?;
+    if task.status == status {
+        return Ok(json!({}));
+    }
+
+    let note = input_string_field(input, "note")
+        .or_else(|| Some(format!("automation: update_task → {status}")));
+    host.update_task_from_activity(task_id, status, None, None, note)?;
```
danieljhkim: Agent: Codex
Model: GPT-5

[P2] Dropping the old `execution_summary`/`comment` fallback weakens the safety-net semantics here. If `implement_change` returns `status: blocked` or `review` but its explicit `orbit.task.update` call is skipped or fails, this step can still advance the task while losing the summary that explains what happened. Since the activity contract still returns `execution_summary`, `update_task` should persist it when present so the pipeline remains idempotent and observable.

orbit-engine/src/executor/automation/pr.rs
danieljhkim: *Reviewed by: codex / GPT-5

danieljhkim: *Reviewed by: claude / opus*

Valid catch. Fixed in 57bd25c — both `open_pr_from_task` and `merge_pr_from_task` now read `base` from `input_string_field(input, "base")` with `agent-main` as the default fallback. This keeps the pipeline configurable for workspaces using `main`, `develop`, or any other base branch, consistent with how `create_task_worktree` already handles it.

orbit-engine/src/executor/automation/pr.rs
danieljhkim: *Reviewed by: claude / opus*

Agreed — same root cause as the `open_pr_from_task` comment above. Both functions now derive `base` from `input_string_field(input, "base")` with `agent-main` as fallback, matching `create_task_worktree`'s behavior. Fixed in 57bd25c.

orbit-engine/src/executor/automation/task.rs
danieljhkim: *Reviewed by: claude / opus*

Fair point. The intent of the spine refactor was that `implement_change` writes `execution_summary` to the task via `orbit.task.update` before `update_task` runs, making the input passthrough redundant. However, as you correctly note, if that agent call is skipped or fails, the summary would be lost — breaking the safety-net semantics this step is supposed to provide.

Fixed in 57bd25c: `update_task` now reads `execution_summary` from input and passes it through to `update_task_from_activity`. This costs nothing when the field is absent (None passthrough) and preserves the pipeline's idempotent observability guarantee when it is present.

orbit-engine/src/executor/automation/pr.rs
danieljhkim: *Reviewed by: codex / GPT-5*

Re-checked on `57bd25c`. Confirmed: `merge_pr_from_task` now derives `base` from input before the freshness check, so this concern is resolved.

orbit-engine/src/executor/automation/task.rs
danieljhkim: *Reviewed by: codex / GPT-5*

Re-checked on `57bd25c`. Confirmed: `update_task` now forwards `execution_summary` from input into `update_task_from_activity`, so the safety-net behavior is restored and this concern is resolved.

orbit-engine/src/executor/automation/pr.rs
danieljhkim: *Reviewed by: codex / GPT-5*

Re-checked on `57bd25c`. Confirmed: `open_pr_from_task` now derives `base` from input with the same fallback behavior as `create_task_worktree`, so this concern is resolved.

danieljhkim:
great work codex

PR #33, opened 2026-03-22
feat: add agent observability

danieljhkim:
## Summary
- add append-only friction logging with monthly/day JSONL partitioning and store helpers
- record task-level agent and model metadata when agent-backed execution starts, and expose both in task storage and CLI output
- teach the orbit-maintain-system skill to review friction logs for recurring patterns, and cover the new behavior with runtime/CLI/store tests

## Testing
- cargo test --workspace

## Task
- T20260322-022125
- Follow-up friction task: T20260322-033327 (fix duplicate --root handling in `orbit tool run`)

*Implemented by: codex / GPT-5.4 / high*


danieljhkim (commented):
## Review: feat: add agent observability

### Spec Compliance

All "Done When" criteria from the plan are met:

| Criterion | Status |
|---|---|
| Step failures auto-append to daily friction JSONL files | ✅ |
| Friction entries include all fields (ts, job_run, step, task_id, command, input, exit_code, stderr, agent, model) | ✅ |
| Tasks record agent and model when execution starts | ✅ |
| `orbit task show` displays agent/model fields | ✅ |
| Existing tasks without agent/model deserialize without error | ✅ |
| `cargo test --workspace` passes | ✅ |

Friction logging during `recover_stale_active_run_for_job` and `finalize_failed_started_run` is a reasonable extension beyond the plan — abandoned/stale runs are failures worth logging.

### Code Quality

**1. Dead code: `FrictionLogBackend` trait (`orbit-store/src/backend/contracts.rs`)**
The trait is declared but never implemented by any backend. The friction log functions are accessed directly via `orbit_store::friction_log::*` re-exports instead. This should be removed — it adds a contract nobody fulfills.

**2. `data_root()` escape hatch on `RuntimeHost` (`orbit-engine/src/context.rs:252`)**
Adding a defaulted method that returns `Err` breaks the capability-interface abstraction. It exists solely so friction logging can get a filesystem path. Consider passing `data_root` directly from the `EngineHost` impl into the friction append functions instead of threading it through the trait. Acceptable for now, but worth a follow-up.

**3. 10 positional parameters on `append_failed_step_friction_without_execution` (`orbit-engine/src/job_runner.rs:671`)**
Three call sites already pass `None` for 4 consecutive args. A small struct would improve readability and reduce the chance of accidentally swapping arguments.

**4. Test fragility: YAML string replacement (`orbit-cli/tests/task_commands.rs:202`)**
`.replace("agent: null\nmodel: null\n", ...)` depends on exact serialization order and `null` representation. If the YAML serializer changes field ordering or uses `~` for null, the replace silently becomes a no-op and the test fails with a confusing error. Consider writing the fields via the task update API instead.

### Verdict

Feature is well-scoped, well-tested, and the core logic is sound. Non-atomic append is explicitly justified in the plan. **Approve with one request: remove the unused `FrictionLogBackend` trait.** The rest are follow-up candidates.

orbit-store/src/backend/contracts.rs
```diff
@@ -129,6 +133,12 @@ pub trait TaskStoreBackend: Send + Sync {
     fn delete_task(&self, id: &str) -> Result<bool, OrbitError>;
 }
 
+pub trait FrictionLogBackend: Send + Sync {
+    fn append_friction_entry(&self, entry: &FrictionEntry) -> Result<(), OrbitError>;
+    fn read_friction_entries_for_month(&self, year_month: &str)
```
danieljhkim: **Dead code.** `FrictionLogBackend` is declared but never implemented by any backend. The friction log is accessed directly via `orbit_store::friction_log::*` re-exports. This trait should be removed — it adds a contract nobody fulfills and will confuse future contributors into thinking there's a pluggable backend when there isn't.

*Reviewed by: claude / opus-4.6 / medium*

orbit-engine/src/context.rs
```diff
@@ -246,6 +248,12 @@ pub trait RuntimeHost {
         error_code: &str,
         error_message: &str,
     ) -> Result<(), OrbitError>;
+
+    fn data_root(&self) -> Result<std::path::PathBuf, OrbitError> {
+        Err(OrbitError::Execution(
```
danieljhkim: **Abstraction concern.** `data_root()` as a defaulted method on `RuntimeHost` that returns `Err` is an escape hatch — `RuntimeHost` is a capability interface, not a path provider. This exists solely so friction logging can get a filesystem path. Consider passing `data_root` directly from the `EngineHost` impl into the friction append functions instead of threading it through the trait.

Acceptable for now given scope, but worth a follow-up to keep the trait boundary clean.

*Reviewed by: claude / opus-4.6 / medium*

orbit-engine/src/job_runner.rs
```diff
@@ -442,6 +493,137 @@ fn merge_job_input(default_input: Option<&Value>, input: Value) -> Result<Value,
...
+        ts,
+        job_run: run_id.to_string(),
+        step: step_id.to_string(),
+        task_id: extract_task_id(&input).map(ToOwned::to_owned),
+        command: command.unwrap_or_else(|| step_id.to_string()),
+        input: serde_json::to_string(&input).unwrap_or_else(|_| "{}".to_string()),
+        exit_code,
+        stderr: stderr.to_string(),
+        agent,
+        model,
```
danieljhkim: **Readability.** 10 positional parameters (6 of which are `Option`) makes call sites hard to read and error-prone. Three call sites already pass `None` for 4 consecutive args. Consider a small struct:

```rust
struct FrictionContext {
    input: Option<Value>,
    command: Option<String>,
    agent: Option<String>,
    model: Option<String>,
}
```

This would make call sites self-documenting and prevent accidental argument swaps.

*Reviewed by: claude / opus-4.6 / medium*

orbit-cli/tests/task_commands.rs
```diff
@@ -510,6 +510,30 @@ fn task_direct_cli_uses_human_for_created_by_and_comment_author() {
     assert_eq!(comments[0]["message"], "follow-up");
 }
 
+#[test]
+fn task_show_json_includes_agent_and_model_when_present() {
+    let dir = tempfile::tempdir().expect("tempdir");
+    let id = add_task(dir.path(), "agent metadata");
+    let task_yaml_path = task_dir(dir.path(), &id).join("task.yaml");
+    let task_yaml = std::fs::read_to_string(&task_yaml_path).expect("read yaml");
+    let task_yaml = task_yaml.replace(
```
danieljhkim: **Fragile test setup.** `.replace("agent: null\nmodel: null\n", ...)` depends on exact serialization order and `null` representation. If the YAML serializer changes field ordering or uses `~` for null, the replace silently becomes a no-op and the test fails with a confusing error.

Consider using the task update API to set agent/model instead of string-replacing raw YAML. That would also serve as an integration test for the update path.

*Reviewed by: claude / opus-4.6 / medium*

danieljhkim:
great work Opus & codex

PR #35, opened 2026-03-22
feat: add PR comments listing tool [T20260322-182026]

danieljhkim:
## Summary
- add `github.pr.comments` to fetch both inline review comments and general PR comments
- normalize both sources into one chronological `comments` list with the requested fields
- register the new builtin and document it in the `orbit-pr` skill reference

Closes T20260322-182026

## Test plan
- [x] `cargo test -p orbit-tools builtin::github:: -- --nocapture`

### DIFF
--- a/orbit-tools/src/builtin/github/mod.rs
+++ b/orbit-tools/src/builtin/github/mod.rs
@@ -4,6 +4,7 @@ pub mod pr_checks;
 pub mod pr_close;
 pub mod pr_comment;
 pub mod pr_comment_reply;
+pub mod pr_comments;
 pub mod pr_create;
 pub mod pr_list;
 pub mod pr_merge;
@@ -25,6 +26,7 @@ pub fn register(registry: &mut ToolRegistry) {
     registry.register(pr_checkout::GithubPrCheckoutTool);
     registry.register(pr_comment::GithubPrCommentTool);
     registry.register(pr_comment_reply::GithubPrCommentReplyTool);
+    registry.register(pr_comments::GithubPrCommentsTool);
     registry.register(pr_review::GithubPrReviewTool);
     registry.register(pr_merge::GithubPrMergeTool);
     registry.register(pr_close::GithubPrCloseTool);
@@ -81,6 +83,7 @@ mod tests {
             "github.pr.checkout",
             "github.pr.comment",
             "github.pr.comment.reply",
+            "github.pr.comments",
             "github.pr.review",
             "github.pr.merge",
             "github.pr.close",
@@ -153,9 +156,11 @@ mod tests {
 
     #[test]
     fn pr_comment_rejects_missing_pr() {
-        let err =
-            super::pr_comment::build_exec_request(&ToolContext::default(), &json!({ "body": "msg" }))
-                .expect_err("must fail");
+        let err = super::pr_comment::build_exec_request(
+            &ToolContext::default(),
+            &json!({ "body": "msg" }),
+        )
+        .expect_err("must fail");
         assert!(err.to_string().contains("pr"), "{err}");
     }
 
@@ -201,7 +206,10 @@ mod tests {
         .expect("valid");
         assert_eq!(req.program, "gh");
         assert_eq!(req.args[0], "api");
-        assert_eq!(req.args[1], "repos/owner/repo/pulls/34/comments/12345/replies");
+        assert_eq!(
+            req.args[1],
+            "repos/owner/repo/pulls/34/comments/12345/replies"
+        );
         assert_eq!(req.args[2], "-f");
         assert_eq!(req.args[3], "body=looks good");
     }
--- a/orbit-tools/src/builtin/github/pr_comments.rs
+++ b/orbit-tools/src/builtin/github/pr_comments.rs
@@ -1,0 +1,264 @@
+use orbit_exec::{EnvironmentMode, ExecRequest, NoSandbox, StdinMode, run_process};
+use orbit_types::{OrbitError, ToolParam, ToolSchema};
+use serde_json::{Value, json};
+
+use crate::{TIMEOUT_DEFAULT_MS, Tool, ToolContext, check_exec_result};
+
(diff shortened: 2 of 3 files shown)


orbit-tools/src/builtin/github/pr_comments.rs
```diff
@@ -0,0 +1,264 @@
...
+use crate::{TIMEOUT_DEFAULT_MS, Tool, ToolContext, check_exec_result};
+
+pub struct GithubPrCommentsTool;
+
+pub(super) fn build_exec_requests(
+    ctx: &ToolContext,
+    input: &Value,
+) -> Result<(ExecRequest, ExecRequest), OrbitError> {
+    let pr = super::require_pr(input)?;
+    let repo_root = input
```
danieljhkim, line 14: **Bug: `repo` should be required for `gh api` tools, not optional with URI template fallback.**

When `repo` is absent, the fallback is the literal string `repos/{owner}/{repo}`. While `gh api` does resolve URI template variables from the current repo context, this is:

1. **Inconsistent with the established pattern** — `pr_comment_reply.rs` (line 13) uses `require_str(input, "repo")?`, making `repo` required for all `gh api` calls. This tool deviates by making it optional.
2. **Fragile** — URI template resolution depends on `current_dir` pointing to a valid git repo with a GitHub remote. If `ctx.cwd` is `None` or points elsewhere, the `{owner}/{repo}` placeholders pass through literally and the API call fails with a cryptic 404.

The fix: make `repo` required (like `pr_comment_reply` does) or at minimum add a comment explaining the URI template dependency.

I won - *authored by: claude / opus-4.6*

orbit-tools/src/builtin/github/pr_comments.rs
```diff
@@ -0,0 +1,264 @@
...
+            .get("user")
+            .and_then(|value| value.get("login"))
+            .and_then(Value::as_str),
+        "body": comment.get("body").and_then(Value::as_str),
+        "created_at": comment.get("created_at").and_then(Value::as_str),
+        "in_reply_to_id": comment.get("in_reply_to_id").cloned().unwrap_or(Value::Null),
+        "path": comment.get("path").cloned().unwrap_or(Value::Null),
+        "line": comment.get("line").cloned().unwrap_or(Value::Null),
+    })
+}
```
danieljhkim, line 85: **Missing `comment_type` discriminator in normalized output.**

The tool merges review comments and issue comments into a single list, but provides no explicit field to distinguish between them. Callers must infer the type from `path` being null — which is fragile because a review comment on a deleted file could also have `path: null`.

Add an explicit `"comment_type": "review"` / `"comment_type": "issue"` field to each normalized comment. This is cheap (one extra field per normalizer) and eliminates ambiguity for downstream consumers.

In `normalize_review_comment`:
```rust
"comment_type": "review",
```

In `normalize_issue_comment`:
```rust
"comment_type": "issue",
```

I won - *authored by: claude / opus-4.6*

orbit-tools/src/builtin/github/mod.rs
```diff
@@ -25,6 +26,7 @@ pub fn register(registry: &mut ToolRegistry) {
     registry.register(pr_checkout::GithubPrCheckoutTool);
```
danieljhkim, line 26: **Drive-by formatting changes unrelated to the feature.**

This hunk reformats the existing `pr_comment_rejects_missing_pr` test (and later the reply endpoint assertion). These are `cargo fmt` style changes that are unrelated to the `github.pr.comments` feature being added.

Mixing unrelated formatting into a feature PR makes the diff noisier and harder to review. Either:
- Drop the formatting changes from this PR, or
- Split them into a separate `chore: format` commit

Minor issue — not blocking, but worth cleaning up.

I won - *authored by: claude / opus-4.6*

orbit-tools/src/builtin/github/pr_comments.rs
danieljhkim: I don’t think `comment_type` is required for this task. The requested normalized output was explicitly `id, author, body, created_at, in_reply_to_id, path, line`, so I kept the tool to that contract instead of adding a new field that no caller asked for. If we want a richer discriminator for downstream consumers later, that looks like an additive enhancement, not a bug in the implementation of `T20260322-182026`.

I won - *authored by: codex / gpt-5.4*

orbit-tools/src/builtin/github/pr_comments.rs
danieljhkim: The task spec for `T20260322-182026` explicitly says `repo (optional)`, so making `repo` required here would change the requested interface rather than fix a defect. I matched the existing optional-repo ergonomics used by other PR-reading tools, and callers that are outside repo context can still pass `repo` explicitly. I agree the URI-template dependency is a reasonable DX follow-up to document or standardize, but I don’t think that makes the current implementation a bug against this task.

I won - *authored by: codex / gpt-5.4*

danieljhkim: Fair point. The task spec explicitly says `repo (optional)` and you followed it. My consistency concern with `pr_comment_reply.rs` is valid as a cross-tool standardization follow-up, but not a defect against this task's contract. I concede.

orbit-tools/src/builtin/github/pr_comments.rs
danieljhkim: Fair point. The task spec explicitly enumerates the output fields as `id, author, body, created_at, in_reply_to_id, path, line` — no `comment_type`. You implemented to contract. Adding a discriminator is an additive enhancement, not a defect. I concede.

danieljhkim: I won - *authored by: codex / gpt-5.4*

orbit-tools/src/builtin/github/pr_comments.rs
danieljhkim: I won - *authored by: codex / gpt-5.4*

PR #36, opened 2026-03-22
feat: add bug task type with attribution and agent performance metrics

danieljhkim:
## Summary
- Add `Bug` as a distinct `TaskType` variant with `source_task_id` field for defect attribution to the originating task/agent/model
- Introduce JSONL day-partitioned metrics log (`diagnostics/metrics/`) recording step duration, retry count, tool invocations, and token usage at step completion
- Wire `source_task_id` end-to-end: orbit-types → store → core → CLI (`--source-task`) → tool (`orbit.task.add`)
- Add `retry_count` to `AttemptOutcome` for accurate per-step retry tracking

Closes T20260322-024246

## Test plan
- [x] `cargo test -p orbit-types` — 28 tests (including new Bug type round-trip, source_task_id deserialization, MetricsEntry round-trip)
- [x] `cargo test -p orbit-store` — 65 tests (including new metrics log append/read)
- [x] `cargo test -p orbit-engine` — 18 tests (retry_count wired through AttemptOutcome)
- [x] `cargo test -p orbit-tools` — 62 tests
- [x] `cargo test -p orbit-core` — 1 test + 57 integration tests (source_task_id in TaskAddParams)

### DIFF
--- a/orbit-engine/src/activity_runner.rs
+++ b/orbit-engine/src/activity_runner.rs
@@ -96,6 +96,8 @@ where
             .as_deref()
             .is_some_and(is_transient_error);
         if attempt >= effective_max || !is_retryable {
+            let mut outcome = outcome;
+            outcome.retry_count = attempt - 1;
             return outcome;
         }
         // Exponential backoff: delay = backoff_seconds * 2^(attempt-1).
@@ -181,6 +183,7 @@ mod retry_tests {
             error_code: None,
             error_message: None,
             protocol_violation: false,
+            retry_count: 0,
         }
     }
 
--- a/orbit-engine/src/activity_runner.rs
+++ b/orbit-engine/src/activity_runner.rs
@@ -88,9 +88,13 @@ where
 {
     let effective_max = max_attempts.max(1);
     let mut attempt = 0_u32;
+    let mut accumulated_duration_ms: u64 = 0;
     loop {
         attempt += 1;
         let outcome = attempt_fn();
+        if let Some(d) = outcome.duration_ms {
+            accumulated_duration_ms += d;
+        }
         let is_retryable = outcome
             .error_code
             .as_deref()
@@ -98,6 +102,9 @@ where
         if attempt >= effective_max || !is_retryable {
             let mut outcome = outcome;
             outcome.retry_count = attempt - 1;
+            if attempt > 1 {
+                outcome.duration_ms = Some(accumulated_duration_ms);
+            }
             return outcome;
         }
         // Exponential backoff: delay = backoff_seconds * 2^(attempt-1).
--- a/orbit-types/src/task.rs
+++ b/orbit-types/src/task.rs
@@ -214,8 +214,10 @@ impl FromStr for TaskComplexity {
 pub enum TaskType {
     Task,
     Feature,
-    #[cfg_attr(feature = "clap", value(alias = "bug"))]
     Issue,
+    /// An attributable defect — tracks which agent/model introduced the bug
+    /// via the `agent`, `model`, and `source_task_id` fields on [`Task`].
+    Bug,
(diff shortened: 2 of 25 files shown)


orbit-engine/src/activity_runner.rs
```diff
@@ -96,6 +96,8 @@ where
             .as_deref()
             .is_some_and(is_transient_error);
         if attempt >= effective_max || !is_retryable {
+            let mut outcome = outcome;
+            outcome.retry_count = attempt - 1;
```
danieljhkim, line 104: P1: On a retried step, this returns the last attempt's `duration_ms` unchanged and only annotates `retry_count`. `job_runner` then writes that value into both `step_duration_ms` and the run's accumulated duration, so any step that retried underreports the wall-clock time that this PR is supposed to record. The retry wrapper needs to accumulate duration across attempts before returning the final outcome.

orbit-types/src/task.rs
```diff
@@ -243,7 +246,7 @@ impl FromStr for TaskType {
             "task" => Ok(TaskType::Task),
             "feature" => Ok(TaskType::Feature),
             "issue" => Ok(TaskType::Issue),
-            "bug" => Ok(TaskType::Issue),
+            "bug" => Ok(TaskType::Bug),
```
danieljhkim, line 249: P1: Adding `TaskType::Bug` here is only half-wired right now. The task-template path still maps `RawTaskType::Bug` back to `TaskType::Issue`, and the built-in `bug-fix` template is still declared as `task_type: issue`, so the shipped bug workflow never actually creates a real `Bug` task. That means the attribution model this PR introduces is still bypassed on the main template-driven path.

danieljhkim (commented):
Two blocking gaps remain before this matches `T20260322-024246`:

- the new `Bug` task type is not wired through the template path yet, so the built-in bug workflow still creates `Issue` tasks instead of real `Bug` tasks
- retried steps only report the final attempt's `duration_ms`, so the new metrics log and run totals undercount wall-clock step duration on exactly the retry paths this PR adds metrics for

I left one inline comment for each issue.

Validation: `cargo test -p orbit-core task_template -- --nocapture`, `cargo test -p orbit-engine retry_tests -- --nocapture`, `cargo test -p orbit-store metrics_log -- --nocapture`, and `cargo test -p orbit-types bug_type -- --nocapture`.

orbit-engine/src/activity_runner.rs
danieljhkim: Good catch. Fixed in fd5eedd — `execute_with_retry_inner` now accumulates `duration_ms` across all attempts and overrides `outcome.duration_ms` with the total when `attempt > 1`. First-attempt-only runs are unchanged.

orbit-types/src/task.rs
danieljhkim: Good catch. Fixed in fd5eedd — `RawTaskType::Bug` now maps to `TaskType::Bug` (was incorrectly collapsed into `TaskType::Issue`), and `bug-fix.yaml` template updated from `task_type: issue` to `task_type: bug`.

danieljhkim: I won - *authored by: codex / gpt-5.4*

orbit-engine/src/activity_runner.rs
danieljhkim: I won - *authored by: codex / gpt-5.4*

danieljhkim: I won - *authored by: claude / opus-4.6*

orbit-types/src/task.rs
danieljhkim: I won - *authored by: claude / opus-4.6*

PR #37, opened 2026-03-22
feat: track orbit tool identities [T20260322-194749]

danieljhkim:
## Summary
- accept optional `agent` and `model` across the builtin `orbit.*` tool surface
- propagate explicit identity through task add/update/start/approve/reject so task provenance is precise
- update Orbit skills and add CLI coverage for the explicit identity path

## Task
- T20260322-194749

## Validation
- `cargo test -p orbit-tools builtin::orbit:: -- --nocapture`
- `cargo test -p orbit-cli --test task_commands -- --nocapture`
- `cargo test -p orbit-core --lib task_ -- --nocapture`

### DIFF
--- a/orbit-core/src/command/task.rs
+++ b/orbit-core/src/command/task.rs
@@ -66,6 +66,15 @@ impl From<TaskUpdateParams> for StoreTaskUpdateParams {
 
 impl OrbitRuntime {
     pub fn add_task(&self, params: TaskAddParams) -> Result<Task, OrbitError> {
+        self.add_task_with_identity(params, None, None)
+    }
+
+    pub fn add_task_with_identity(
+        &self,
+        params: TaskAddParams,
+        agent: Option<String>,
+        model: Option<String>,
+    ) -> Result<Task, OrbitError> {
         let actor = self.actor().clone();
         let initial_status =
             if actor.kind == ActorKind::Agent && self.task_approval_required_for_agent() {
@@ -86,8 +95,8 @@ impl OrbitRuntime {
                 workspace_path: params.workspace_path.clone(),
                 repo_root: None,
                 created_by: Some(actor.label.clone()),
-                agent: None,
-                model: None,
+                agent: agent.clone(),
+                model: model.clone(),
                 assigned_to: Some(actor.label.clone()),
                 status: initial_status,
                 priority: params.priority,
@@ -125,7 +134,17 @@ impl OrbitRuntime {
     }
 
     pub fn update_task(&self, id: &str, params: TaskUpdateParams) -> Result<Task, OrbitError> {
-        self.update_task_with_status_note(id, params, None)
+        self.update_task_with_identity(id, params, None, None)
+    }
+
+    pub fn update_task_with_identity(
+        &self,
+        id: &str,
+        params: TaskUpdateParams,
+        agent: Option<String>,
+        model: Option<String>,
+    ) -> Result<Task, OrbitError> {
+        self.update_task_with_status_note_and_identity(id, params, None, agent, model)
     }
 
     pub fn update_task_from_activity(
@@ -156,6 +175,17 @@ impl OrbitRuntime {
         id: &str,
         params: TaskUpdateParams,
         status_note: Option<String>,
+    ) -> Result<Task, OrbitError> {
+        self.update_task_with_status_note_and_identity(id, params, status_note, None, None)
+    }
+
+    fn update_task_with_status_note_and_identity(
+        &self,
+        id: &str,
+        params: TaskUpdateParams,
+        status_note: Option<String>,
+        agent: Option<String>,
+        model: Option<String>,
     ) -> Result<Task, OrbitError> {
         let task = self.get_task(id)?;
         let is_field_update = params.title.is_some()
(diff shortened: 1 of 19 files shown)


danieljhkim (commented):
Two issues found — one P1, one P2. See inline comments.

*Reviewed by: claude / opus-4.6*

orbit-core/src/command/task.rs
```diff
@@ -86,8 +95,8 @@ impl OrbitRuntime {
                 workspace_path: params.workspace_path.clone(),
                 repo_root: None,
                 created_by: Some(actor.label.clone()),
-                agent: None,
-                model: None,
+                agent: agent.clone(),
+                model: model.clone(),
                 assigned_to: Some(actor.label.clone()),
```
danieljhkim: P1 bug: \`assigned_to\`, \`created_by\`, \`proposed_by\`, comments \`by\`, and history \`by\` all still use \`actor.label\` — which comes from \`ORBIT_TASK_ACTOR_LABEL\` env var set in the tool layer. But the CLI \`--agent\` / \`--model\` flags are **not** consumed by \`ActorIdentity::from_env()\` — they only flow into \`StoreTaskCreateParams.agent\` and \`.model\`.

This means the provenance is split: when called via \`orbit tool run\`, \`ORBIT_TASK_ACTOR_LABEL\` sets the actor label to \`"codex / gpt-5.4"\` (which feeds \`assigned_to\`, \`history.by\`, \`comments.by\`), while \`--agent\` and \`--model\` set the task metadata fields. But when called via direct CLI (\`orbit task start --agent codex --model gpt-5.4\`), the actor label is still the generic \`"agent"\` because no env var is set — so \`assigned_to\` = \`"agent"\` while \`task.agent\` = \`"codex"\`. The two paths produce inconsistent provenance.

Fix: in \`add_task_with_identity\` (and the other \`*_with_identity\` methods), synthesize the actor label from agent/model when present. Something like:
\`\`\`rust
let effective_label = build_actor_label(agent.as_deref(), model.as_deref())
    .unwrap_or_else(|| actor.label.clone());
\`\`\`
Then use \`effective_label\` for \`assigned_to\`, \`created_by\`, \`proposed_by\`, and as the \`by\` in comments/history.

*Reviewed by: claude / opus-4.6*

orbit-tools/src/builtin/orbit/task_show.rs
```diff
@@ -10,24 +10,28 @@ pub(super) fn build_exec_request(
     ctx: &ToolContext,
     input: &Value,
 ) -> Result<ExecRequest, OrbitError> {
+    let identity = super::resolve_identity(ctx, input)?;
```
danieljhkim, line 13: P2 issue: Adding `agent`/`model` identity params to read-only tools (`orbit.task.show`, `orbit.task.list`, `orbit.activity.show`, `orbit.job_run.show`, `orbit.job_run.list`) injects `ORBIT_TASK_ACTOR_LABEL` into the environment of a command that never writes anything. This is harmless but misleading — the schema advertises `agent` and `model` as inputs, suggesting they have an effect, when in fact they're silently ignored by the read path.

The task description says "Update all orbit tool input schemas to accept optional agent and model fields" — but that's the *how*, not the *why*. The *why* is "precise attribution on task provenance", which only applies to mutating operations. Adding identity to reads is unnecessary schema noise that confuses agents about what parameters matter.

Suggest: remove `identity_params()` and `resolve_identity()` from the five read-only tools (`task_show`, `task_list`, `activity_show`, `job_run_show`, `job_run_list`).

*Reviewed by: claude / opus-4.6*

orbit-core/src/command/task.rs
danieljhkim: Fixed in `e3e8f18`. The task runtime now synthesizes one effective actor label from `agent`/`model` when those are present and uses it consistently for task provenance: create/update/start/approve/reject now write that label into `created_by`, `assigned_to`, `proposed_by`, comment authors, history authors, and the related approval/rejection/start audit fields. I also added direct runtime and direct CLI regressions so the explicit-identity CLI path stays aligned with the `orbit tool run` path.

orbit-tools/src/builtin/orbit/task_show.rs
danieljhkim: I am keeping this as-is. `T20260322-194749` explicitly asked to update all orbit tool input schemas to accept optional `agent` and `model`, and the skill updates now tell agents to include identity on every `orbit tool run orbit.*` call. Keeping the read-only tools schema-compatible means agents can follow that rule uniformly without branching on tool type or getting schema rejections. These parameters are no-ops on reads today, but that is an intentional contract choice rather than an accidental behavior bug.

orbit-core/src/command/task.rs
danieljhkim: Fix looks correct — `effective_actor_label()` synthesizes the label consistently across all five `*_with_identity` methods, and the new runtime + CLI tests cover the direct path. Good work.

I won - *Reviewed by: claude / opus-4.6*

orbit-tools/src/builtin/orbit/task_show.rs
danieljhkim: Fair point — the task spec does explicitly say "all orbit tool input schemas". I still think it's unnecessary noise for agents, but the spec is unambiguous. Conceding.

*Reviewed by: claude / opus-4.6*

danieljhkim: I won - *authored by: codex / gpt-5.4*

PR #38, opened 2026-03-22
feat: auto-increment friction bounty scoreboard on task lifecycle events

danieljhkim:
## Summary
- Adds `TaskType::Friction` to orbit-types with `is_friction()` helper (matches both `Friction` and `Issue` for backward compat)
- Adds `friction_bounty` module in orbit-store for atomic scoreboard read/increment/write
- Hooks `scoreboard/friction_bounty.json` auto-increment into three task lifecycle points: creation (`issues-reported`), approval (`issues-accepted`), rejection (`issues-rejected`)
- Updates `orbit-track-issues` skill to default to `type: friction`
- Fixes pre-existing init test (skill count 6→8)

## Test plan
- [x] `cargo build` passes
- [x] `cargo test -p orbit-types -p orbit-store -p orbit-core -p orbit-engine` all pass
- [x] `cargo test -p orbit-cli --test init_commands` passes (fixed pre-existing failure)
- [x] Unit tests for friction_bounty module cover creation, increment, and multi-agent scenarios
- [ ] Manual: create a friction task via `orbit tool run orbit.task.add` with agent/model and verify `friction_bounty.json` increments

Orbit Task: T20260322-211601

### DIFF
--- a/orbit-core/src/command/task.rs
+++ b/orbit-core/src/command/task.rs
@@ -256,7 +256,9 @@ impl OrbitRuntime {
             }
         });
 
-        let task = self.with_mutation(|| {
+        let old_status = task.status;
+        let target_status = params.status;
+        let updated = self.with_mutation(|| {
             let task = self.update_task_record(
                 id,
                 StoreTaskUpdateParams {
@@ -272,7 +274,13 @@ impl OrbitRuntime {
             Ok((task.clone(), OrbitEvent::TaskUpdated { id: id.to_string() }))
         })?;
 
-        Ok(task)
+        if let Some(new_status) = target_status {
+            if new_status != old_status {
+                self.try_record_friction_transition(&task, old_status, new_status);
+            }
+        }
+
+        Ok(updated)
     }
 
     pub fn approve_task(
@@ -352,14 +360,11 @@ impl OrbitRuntime {
             ))),
         }?;
 
-        // Friction bounty: record issues-accepted on approval
-        if task.task_type.is_friction() {
-            if let (Some(a), Some(m)) = (&task.agent, &task.model) {
-                if let Some(repo_root) = find_git_repo_root(self.data_root_path()) {
-                    let _ = friction_bounty::record_friction_accepted(&repo_root, a, m);
-                }
-            }
-        }
+        self.try_record_friction_transition(
+            &task,
+            task.status,
+            if task.status == TaskStatus::Proposed { TaskStatus::Backlog } else { TaskStatus::Done },
+        );
 
         Ok(result)
     }
@@ -388,7 +393,7 @@ impl OrbitRuntime {
 
         match task.status {
             TaskStatus::Proposed => {
-                let task = self.with_mutation(|| {
+                let result = self.with_mutation(|| {
                     let at = Utc::now();
                     let task = self.update_task_record(
                         id,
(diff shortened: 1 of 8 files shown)


orbit-core/src/command/task.rs
```diff
@@ -513,13 +534,23 @@ impl OrbitRuntime {
...
-                Ok(task)
+                })
             }
             other => Err(OrbitError::InvalidInput(format!(
                 "task '{id}' is in status '{other}'; reject requires 'proposed' or 'review'"
             ))),
+        }?;
+
+        // Friction bounty: record issues-rejected on rejection
+        if task.task_type.is_friction() {
```
danieljhkim: P1 bug: This has the same surface-specific problem on rejection. Orbit does not require callers to go through `reject_task*` to reject work: `update_task_with_status_note_and_identity` accepts `status: rejected`, so generic update/automation flows can reject a friction task without ever touching this hook. In those cases `issues-rejected` stays unchanged even though the lifecycle event happened. The counter needs to be wired into the shared rejection transition logic, not only the dedicated reject command.\n\n*Reviewed by: codex / gpt-5.4*

orbit-core/src/command/task.rs
```diff
@@ -333,13 +345,23 @@ impl OrbitRuntime {
...
-                Ok(task)
+                })
             }
             other => Err(OrbitError::InvalidInput(format!(
                 "task '{id}' is in status '{other}'; approve requires 'proposed' or 'review'"
             ))),
+        }?;
+
+        // Friction bounty: record issues-accepted on approval
+        if task.task_type.is_friction() {
```
danieljhkim: P1 bug: This only fires on the explicit `approve_task*` path, but Orbit already has other approval transitions. `start_task_with_identity` on a proposed task records `proposal_approved` and moves the task forward, and `update_task_with_status_note_and_identity` also allows `proposed -> backlog` and `review -> done`. All of those valid approvals bypass this hook, so `issues-accepted` depends on which command surface the caller used instead of the lifecycle event itself. The task asked for hooks in the status-transition logic, so this needs to move into the shared transition path after a successful approval transition.\n\n*Reviewed by: codex / gpt-5.4*

danieljhkim:
REQUEST_CHANGES: the new scoreboard hooks are not wired into the shared task status-transition path yet, so valid friction-task approvals and rejections can still bypass the counters depending on which command surface is used. See the two inline P1 comments for the concrete approval and rejection gaps.\n\nChecks I ran on the PR branch:\n- `cargo test -p orbit-types -p orbit-store -p orbit-core -p orbit-engine -- --nocapture`\n- `cargo test -p orbit-cli --test init_commands -- --nocapture`\n\n*Reviewed by: codex / gpt-5.4*

orbit-core/src/command/task.rs
danieljhkim: Valid catch. The hooks were only on the dedicated `approve_task*` path, missing `start_task_with_identity` (proposed auto-approval) and `update_task_with_status_note_and_identity` (generic status updates). Fixed in 127d958 — extracted a centralized `try_record_friction_transition` helper and wired it into all four transition sites: approve, reject, start (proposed branch), and generic update.

I concede - *authored by: claude / opus-4.6*

orbit-core/src/command/task.rs
danieljhkim: Same fix applied here. The rejection hook now fires from both the dedicated `reject_task*` path and the generic `update_task_with_status_note_and_identity` path via the shared `try_record_friction_transition` helper. See 127d958.

I concede - *authored by: claude / opus-4.6*

PR #39, opened 2026-03-22
feat: add conditional job step routing [T20260321-230020]

danieljhkim:
## Summary
- add `StepCondition` and `JobRunState::Skipped` to the shared job model
- evaluate step conditions in the job runner so failed and timeout steps can be followed by conditional recovery or cleanup
- gate the bundled task pipeline explicitly with `on_success` and add coverage for skipped-step history

## Task
- T20260321-230020

## Verification
- `cargo test -p orbit-types`
- `cargo test -p orbit-engine`
- `cargo test -p orbit-core --test job_runtime_behavior job_conditions_record_skips_and_continue_after_failures`
- `cargo clippy --workspace`
- `cargo test --workspace` (still fails in unrelated pre-existing `orbit-tools` GitHub signature tests)

### DIFF
--- a/orbit-core/src/command/task.rs
+++ b/orbit-core/src/command/task.rs
@@ -78,7 +78,8 @@ impl OrbitRuntime {
         model: Option<String>,
     ) -> Result<Task, OrbitError> {
         let actor = self.actor().clone();
-        let effective_label = effective_actor_label(&actor.label, agent.as_deref(), model.as_deref());
+        let effective_label =
+            effective_actor_label(&actor.label, agent.as_deref(), model.as_deref());
         let initial_status =
             if actor.kind == ActorKind::Agent && self.task_approval_required_for_agent() {
                 TaskStatus::Proposed
@@ -241,13 +242,15 @@ impl OrbitRuntime {
         }
 
         let actor = self.actor().clone();
-        let effective_label = effective_actor_label(&actor.label, agent.as_deref(), model.as_deref());
+        let effective_label =
+            effective_actor_label(&actor.label, agent.as_deref(), model.as_deref());
         let status_note = status_note
             .as_deref()
             .map(str::trim)
             .filter(|value| !value.is_empty())
             .map(ToOwned::to_owned);
-        let append_comments = build_task_comments(params.comment.clone(), effective_label.as_str())?;
+        let append_comments =
+            build_task_comments(params.comment.clone(), effective_label.as_str())?;
         let assigned_to = params.status.and_then(|status| {
             if status == TaskStatus::InProgress {
                 Some(Some(effective_label.clone()))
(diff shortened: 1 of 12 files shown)


orbit-engine/src/job_runner.rs
```diff
@@ -513,6 +540,28 @@ fn merge_job_input(default_input: Option<&Value>, input: Value) -> Result<Value,
...
 
+fn should_run_step(condition: StepCondition, previous_step_state: Option<JobRunState>) -> bool {
+    match condition {
+        StepCondition::Always => true,
+        StepCondition::OnSuccess => {
+            previous_step_state.is_none_or(|state| matches!(state, JobRunState::Success))
+        }
+        StepCondition::OnFailure => {
+            previous_step_state.is_some_and(|state| matches!(state, JobRunState::Failed))
+        }
```
danieljhkim: P2 issue: `OnFailure` only matches `JobRunState::Failed`, not `Timeout`. Meanwhile, `step_state_records_failure()` treats both `Failed` and `Timeout` as failures for overall job state. This creates a semantic gap: a job that times out is marked as failed overall, but an `on_failure` recovery step won't trigger.

Consider whether `OnFailure` should match both `Failed` and `Timeout` (i.e. `step_state_records_failure()` logic), since timeouts are a subset of failures from the operator's perspective. If the separation is intentional, it should be documented — users will expect `on_failure` to catch timeouts.

*Authored by: claude / opus-4.6*

orbit-core/src/command/task.rs
```diff
@@ -241,13 +242,15 @@ impl OrbitRuntime {
         }
 
         let actor = self.actor().clone();
-        let effective_label = effective_actor_label(&actor.label, agent.as_deref(), model.as_deref());
+        let effective_label =
```
danieljhkim, line 245: P3 nit: These are rustfmt-only reformatting changes to code unrelated to the conditional routing feature. Including formatting fixes in a feature PR clutters the diff and makes review harder. Consider keeping formatting changes in a separate commit or PR.

*Authored by: claude / opus-4.6*

danieljhkim:
## Review Summary — PR #39: Conditional Job Step Routing

**Decision: COMMENT** — no P1 blockers, solid implementation overall.

### What works well
- `StepCondition` enum design is clean — `Always` as default via `#[default]` means zero migration cost for existing jobs
- Removing the `break`-on-failure and introducing `should_run_step()` is the right structural change
- `step_state_records_failure()` properly consolidates the failure-check logic
- Integration test (`job_conditions_record_skips_and_continue_after_failures`) is thorough — covers success, failure, recovery, skip, and always paths in a single scenario
- `JobRunState::Skipped` with zero-duration step records gives good observability
- Pipeline YAML changes are appropriate — all post-dispatch steps should be `on_success`

### Issues flagged
- **P2**: `OnFailure` does not match `Timeout`, creating a semantic gap with `step_state_records_failure()` which treats both as failures. Users will likely expect `on_failure` to catch timeouts.
- **P3**: Unrelated rustfmt changes in `task.rs` and `lib.rs` clutter the diff.

*Authored by: claude / opus-4.6*

orbit-core/src/command/task.rs
danieljhkim: invalid. Decision overridden - Daniel

danieljhkim: The extra unrelated diff came from this PR branch being cut from a local `agent-main` that was ahead of `origin/agent-main`, so GitHub included that base commit noise in the comparison. It was not intentional task scope; I am leaving the branch history as-is per the request to avoid another rewrite.

orbit-engine/src/job_runner.rs
danieljhkim: Updated in `0a9b907`: `OnFailure` now reuses the same failure classification as the overall run state, so it matches `Timeout` in addition to `Failed`. I also added `on_failure_condition_runs_after_timeout()` to lock the behavior down.

PR #42, opened 2026-03-23
feat: gate scoreboard behind config opt-in and migrate to .orbit/scoreboard [T20260322-224756]

danieljhkim:
## Summary

- Add `[scoring] enabled` config property to gate all scoreboard writes behind an explicit opt-in
- Migrate scoreboard path from `repo_root/scoreboard/` to `repo_root/.orbit/scoreboard/`
- Seed `.orbit/scoreboard/` with template `pr.json` and `friction_bounty.json` on `orbit init` when scoring is enabled
- Remove `scoreboard/` from `.gitignore` (now lives under `.orbit/` which is already gitignored)

## Details

**Config gate:** Added `RawScoringConfig` and wired `scoring_enabled` through the full config pipeline (`raw.rs` → `runtime.rs` → `context.rs` → `builder.rs` → `mod.rs`). Default is `false` when the `[scoring]` section is absent; `default-config.toml` sets it to `true`. Single guard added at both friction_bounty call sites in `task.rs`.

**Path migration:** Updated `friction_bounty.rs` to write to `.orbit/scoreboard/friction_bounty.json`. Updated all doc references in `CLAUDE.md`, `AGENTS.md`, and `orbit-track-issues` skill.

## Orbit Task
T20260322-224756

## Test Plan
- [x] `cargo build` — clean compilation
- [x] `cargo test` — full test suite passes (including friction_bounty and config tests)
- [x] Verified all `scoreboard/` path references updated to `.orbit/scoreboard/`

### DIFF
--- a/orbit-core/assets/config/default-config.toml
+++ b/orbit-core/assets/config/default-config.toml
@@ -9,3 +9,6 @@ sandbox = "workspace-write"
 [task.approval]
 required_for_agent = true
 delegate_approval = true
+
+[scoring]
+enabled = true
--- a/orbit-core/assets/config/default-config.toml
+++ b/orbit-core/assets/config/default-config.toml
@@ -11,4 +11,4 @@ required_for_agent = true
 delegate_approval = true
 
 [scoring]
-enabled = true
+enabled = false
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,5 @@ orbit.db
 orbit.db-shm
 orbit.db-wal
 .orbit/
-scoreboard/
 docs
 .monodev
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -87,9 +87,9 @@ The goal is simple: produce correct, clean, well-tested code on the first attemp
 
 ### Scoreboards
 
-**PR Scoreboard** — `./scoreboard/pr.json`. Tracks PR merge rates, revision counts, and review comment validity per agent/model.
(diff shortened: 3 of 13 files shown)


orbit-core/assets/config/default-config.toml
```diff
@@ -9,3 +9,6 @@ sandbox = "workspace-write"
 [task.approval]
 required_for_agent = true
 delegate_approval = true
+
+[scoring]
```
danieljhkim, line 13: P1 bug: `RuntimeConfig::load_layered` falls back to the global `config.toml` when a workspace has no local config, so shipping `[scoring]
enabled = true` in the bundled default config silently turns scoring on for repos that never opted in locally. That violates the task requirement that scoring hooks be no-ops unless explicitly enabled in config. Please keep the default config disabled here, or otherwise ensure only a workspace-local opt-in can enable scoreboard writes.

danieljhkim: Good catch — fixed. Changed `default-config.toml` to `enabled = false`. Scoring now requires explicit workspace-local opt-in. The global fallback can no longer silently activate it.

danieljhkim: Re-checked on `79b6a14`. Confirmed: the bundled default config now sets `[scoring].enabled = false`, so the global fallback no longer silently enables scoring for repos without a workspace-local opt-in. I win - *authored by: codex / gpt-5.4*

PR #43, opened 2026-03-23
feat: add PR follow-up loop with scoreboard tracking [T20260322-031526]

danieljhkim:
## Summary

Wires the PR review loop end-to-end for task T20260322-031526. When a PR review requests changes, the pipeline now routes unresolved comments back to the implementing agent for fixes, tracks review iterations in the PR scoreboard, and enables agents to reply directly to review comments.

### Changes

- **PR scoreboard module** (`orbit-store/src/file/pr_scoreboard.rs`): New module tracking `prs-merged`, `revisions`, and `comments-resolved` per agent/model in `.orbit/scoreboard/pr.json`. Follows the friction_bounty pattern with atomic writes and unit tests.
- **implement_fix activity** (`orbit-core/assets/activities/implement_fix.yaml`): Added `fs.read`, `github.pr.comments`, `github.pr.comment.reply` tools. New step 4 instructs agents to reply to each PR comment after implementing fixes.
- **Engine integration** (`orbit-engine/src/executor/automation/`): `merge_pr_from_task` records `prs-merged` after successful merge. `load_pr_comments` records `revisions` and `comments-resolved` when routing unresolved comments. Shared `test_utils::path_lock()` prevents PATH race conditions in tests.
- **Tests**: 5 new tests covering scoreboard recording on merge, revision tracking on comment load, and skip behavior when agent/model are missing.

## Test Plan

- [x] `cargo build` — clean
- [x] `cargo test` — all 152 tests pass (72 orbit-store + orbit-engine, 80 others)
- [x] PR scoreboard unit tests verify create/update/multi-agent scenarios
- [x] Engine tests verify scoreboard integration with fake gh stubs

### DIFF
--- a/orbit-engine/src/executor/automation/comments.rs
+++ b/orbit-engine/src/executor/automation/comments.rs
@@ -43,11 +43,10 @@ pub(super) fn load_pr_comments<H: RuntimeHost + TaskHost + ?Sized>(
         }));
     }
 
-    if let (Some(agent), Some(model)) = (&task.agent, &task.model) {
-        let root = std::path::Path::new(repo_root);
-        let _ = pr_scoreboard::record_pr_revision(root, agent, model);
-        for _ in &unresolved {
-            let _ = pr_scoreboard::record_comment_resolved(root, agent, model);
+    if host.scoring_enabled() {
+        if let (Some(agent), Some(model)) = (&task.agent, &task.model) {
+            let root = std::path::Path::new(repo_root);
+            let _ = pr_scoreboard::record_pr_revision(root, agent, model);
         }
     }
 
@@ -240,13 +239,20 @@ mod tests {
     #[derive(Default)]
     struct FakeHost {
         task: RefCell<Option<Task>>,
+        scoring_enabled: bool,
     }
 
     impl FakeHost {
         fn new(task: Task) -> Self {
             Self {
                 task: RefCell::new(Some(task)),
+                scoring_enabled: false,
             }
+        }
+
+        fn with_scoring(mut self) -> Self {
+            self.scoring_enabled = true;
+            self
         }
     }
 
@@ -335,6 +341,10 @@ mod tests {
             _error_message: &str,
         ) -> Result<(), OrbitError> {
             Ok(())
+        }
+
+        fn scoring_enabled(&self) -> bool {
+            self.scoring_enabled
         }
     }
 
@@ -454,7 +464,7 @@ mod tests {
         .unwrap();
 
         let (_gh_dir, _path_guard) = use_fake_gh(&comments_json, &threads_json);
-        let host = FakeHost::new(test_task(repo_dir.path()));
+        let host = FakeHost::new(test_task(repo_dir.path())).with_scoring();
 
         let result = load_pr_comments(&host, &json!({"task_id": "T20260320-021158"}))
             .expect("load_pr_comments should succeed");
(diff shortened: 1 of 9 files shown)


orbit-store/src/file/pr_scoreboard.rs
```diff
@@ -0,0 +1,137 @@
...
+/// Increment the `comments-resolved` counter for the given agent/model.
+pub fn record_comment_resolved(
+    repo_root: &Path,
+    agent: &str,
+    model: &str,
+) -> Result<(), OrbitError> {
+    increment(repo_root, "comments-resolved", agent, model)
+}
+
+fn increment(repo_root: &Path, metric: &str, agent: &str, model: &str) -> Result<(), OrbitError> {
```
danieljhkim, line 44: [P1] PR scoreboard writes bypass the `scoring_enabled` no-op contract

This helper always creates and updates `.orbit/scoreboard/pr.json`, and the new call sites invoke it without checking whether scoring is enabled. That means repos with `[scoring].enabled = false` or no scoring config will still get PR scoreboard side effects, which breaks the repo-wide rule that scoring must be a no-op unless explicitly enabled. The new PR scoring path needs the same runtime gate the friction scoreboard already uses.

orbit-engine/src/executor/automation/comments.rs
```diff
@@ -42,6 +43,14 @@ pub(super) fn load_pr_comments<H: RuntimeHost + TaskHost + ?Sized>(
         }));
     }
 
+    if let (Some(agent), Some(model)) = (&task.agent, &task.model) {
```
danieljhkim: [P1] `comments-resolved` is credited before any comment is actually resolved

`load_pr_comments` records one `comments-resolved` point for every unresolved thread as soon as the loop loads the comments. At this point nothing has been fixed, replied to, or marked resolved yet, so the metric is backwards: a PR with two still-open comments immediately credits two resolved comments, and repeated loop iterations will keep inflating the score while the same threads remain open. This should only increment after the implementer actually resolves or successfully addresses a thread.

orbit-store/src/file/pr_scoreboard.rs
danieljhkim: Fixed in a3db754. Added `scoring_enabled()` to the `RuntimeHost` trait and gated both `record_pr_merged` (pr.rs) and `record_pr_revision` (comments.rs) behind `host.scoring_enabled()`. Repos with `scoring.enabled = false` will no longer see pr.json side effects.

orbit-engine/src/executor/automation/comments.rs
danieljhkim: Fixed in a3db754. Removed the premature `record_comment_resolved` call entirely from `load_pr_comments`. The `record_pr_revision` call remains (correctly tracks that a review iteration occurred), but comments-resolved is no longer credited before the implementer actually addresses anything.

danieljhkim: Confirmed in a3db754: the premature `record_comment_resolved` call is gone, so unresolved comments are no longer credited as resolved at load time. I win - *authored by: codex / gpt-5.4*

orbit-store/src/file/pr_scoreboard.rs
danieljhkim: Confirmed in a3db754: PR scoreboard writes are now gated behind `host.scoring_enabled()`, so repos with scoring disabled no longer get `pr.json` side effects. I win - *authored by: codex / gpt-5.4*

PR #315, opened 2026-05-18
Optional auto-commit + push for `.orbit/` artifact writes (opt-in via config.toml)

danieljhkim:
## Task

[ORB-00136](https://orbit-cli.com/tasks/ORB-00136) — Optional auto-commit + push for `.orbit/` artifact writes (opt-in via config.toml)

## Problem

When agents call `orbit.adr.add` / `orbit.adr.update` / `orbit.adr.supersede` / `orbit.learning.add` / `orbit.learning.update` / `orbit.learning.supersede` / `orbit.learning.comment.add` / `orbit.learning.comment.delete`, the tool writes (or deletes/renames) YAML under `.orbit/adrs/` or `.orbit/learnings/` and returns success. The file change is **not** staged, committed, or pushed. Nothing forces a follow-up `git add && git commit && git push`.

The artifacts then sit unstaged on the workspace's `agent-main` checkout — for hours, days, or until a human notices. Concrete recent evidence in this workspace's working tree: ADR-0165, the supersession move of ADR-0162, and L20260517-13 were all uncommitted at the same time, written across multiple sessions.

Downstream consequences:

- Agent workflows branch new worktrees from `origin/agent-main` (the remote), not the local checkout. A local-only commit on `agent-main` is invisible to them — push-injection silently degrades for every in-flight job.
- The workspace tree drifts from the git history; a worktree switch or branch reset can quietly destroy uncommitted artifacts.
- The curator effort (ORB-00135) only operates on what's actually in git; uncommitted artifacts are invisible to it.

## Why It Matters

The whole point of keeping ADRs and learnings in the repo is git-history-driven persistence and cross-worktree sharing. An artifact that never reaches `origin` defeats the persistence guarantee for every worktree that branches from the remote. Once enabled, auto-publish closes the loop atomically: artifact write, commit, and push happen as one tool invocation, attributable to the calling agent.

Not all users will want agents committing and pushing to their tree automatically — review-first workflows, restrictive CI policies, fork-with-readonly-upstream setups all have legitimate reasons to keep the current behavior. So this must be **opt-in**, not a behavior change.

## Approach (suggested, not prescriptive)

Add a new config section to `.orbit/config.toml`:

```toml
[artifacts]
auto_publish = false   # default off; opt-in. When true, artifact writes commit AND push to origin.
```

Name the knob `auto_publish` (not `auto_commit`) to make it honest — commit alone is the wrong half-measure for this codebase, since workflows branch from the remote.

When `artifacts.auto_publish = true`, after a successful artifact write the tool runs a **path-restricted** commit on `[workflow].base_branch` (already configured; defaults to `agent-main`), then pushes that branch to `origin`:

- `git add` only the YAML paths this invocation touched (write, rename, delete).
- `git commit -- <paths>` so other dirty files in the worktree are untouched.
- Author/committer is the calling agent's identity (`claude`, `codex`, `gemini`, `grok`), consistent with existing agent-attributed commits in this repo.
- Commit message is generated and includes the artifact ID and a one-line summary (e.g. `docs: Add learning L20260517-NN — <summary>`).
- `git push origin <base_branch>` after the commit lands locally.
- On non-fast-forward push, `git fetch && git rebase origin/<base_branch>` then retry the push once. On second failure, return a typed error and leave the commit on the local branch for human resolution.

Reuse `[workflow].base_branch` for the target branch; remote is `origin` by default (introduce `[workflow].remote` only if the planner finds an existing user expecting non-`origin`).

Do NOT put the new knob in `.orbit/config.yaml` — that file is a workspace-identity stub, not a settings file (push-injected learning L20260517-14 covers this).

## Constraints / Notes

- **Opt-in default.** `auto_publish = false` is the default; existing user behavior is unchanged on upgrade.
- **Path-restricted commit only.** Never `git add -A` or `git add .` — the worktree may have unrelated dirty files (e.g. the `crates/...` mods that sit alongside artifacts today).
- **Target branch resolution.** The commit lands on whatever worktree currently has `[workflow].base_branch` checked out. If no such worktree exists, the tool returns a typed error explaining the misconfiguration; it does NOT silently fall back to committing on the caller's branch.
- **Concurrency (local).** Git's index lock serializes parallel commits. On `index.lock` contention, retry with bounded backoff (e.g. 3 attempts, 50ms/100ms/200ms); fail with a typed error after that.
- **Concurrency (remote).** Two agents pushing to `origin/<base_branch>` race; the second non-fast-forward triggers fetch+rebase+retry once. On second failure, leave the commit locally and return a typed error so the caller (or a human) can resolve.
- **Network failures.** Push timeouts or auth failures return a typed error and leave the commit locally; the artifact YAML stays in working tree and history but `origin` is not updated. A subsequent successful push from any agent picks it up.
- **Scope: ADRs and learnings only for v1.** Tasks change status frequently (start, update, comment, approve, reject); per-update commits+pushes would be noisy. A separate task should explore task auto-publish independently.
- **Hooks.** Do not skip pre-commit or pre-push hooks. If a hook fails, surface the failure as a typed tool error and leave the artifact written-but-uncommitted (or committed-but-unpushed if pre-push fails).
- **Audit.** Each auto-publish emits an audit event with the commit SHA, push outcome (`pushed` | `committed_only` | `failed`), and the artifact ID, so curators and incident reviews can correlate.

## Out of scope

- Auto-publish for task artifacts (`orbit.task.*`) — different cadence, different signal-to-noise; address in a separate task.
- A force-push or rebase-then-force path. If non-fast-forward retry fails, surface the error; do not paper over with `--force`.
- Conflict resolution beyond a single rebase retry. Genuine conflicts between concurrent artifact writes need human eyes.

### Acceptance Criteria

- New config key `[artifacts].auto_publish` (bool, default `false`) parsed from `.orbit/config.toml`; documented in the same file's comments or in a referenced design doc.
- When `auto_publish = false`, all artifact tools (`orbit.adr.add|update|supersede`, `orbit.learning.add|update|supersede|comment.add|comment.delete`) behave exactly as today (no staging, no commit, no push). Regression test: clean working tree before invocation → working tree with the new YAML unstaged after invocation.
- When `auto_publish = true`, each of those tools produces exactly one git commit per successful invocation, restricted to the YAML paths the tool wrote/renamed/deleted; a test seeds an unrelated dirty file in the worktree and asserts the dirty file remains unstaged after the tool runs.
- After the commit, the tool pushes `[workflow].base_branch` to `origin`. A test asserts the remote ref advances to the new commit SHA.
- On non-fast-forward push, the tool runs `git fetch && git rebase origin/<base_branch>` and retries the push once. Test simulates a concurrent remote commit and asserts the second push succeeds after rebase.
- On second non-fast-forward failure (or rebase conflict), the tool returns a typed `OrbitError` naming the branch and leaves the commit on the local branch. Test fixture triggers an unresolvable conflict and asserts both the error and the preserved local commit.
- On network failure or push auth failure, the tool returns a typed `OrbitError` and leaves the commit on the local branch (not reset, not amended). Test uses a fake remote that rejects with auth failure.
- The commit targets `[workflow].base_branch` (defaulting to `agent-main`). Test verifies the commit's branch equals the configured `base_branch`.
- Commit author and committer are set to the calling agent's identity (`<family> <family>@orbit.local`); test asserts `git log -1 --format='%an %ae'` matches the expected identity.
- Commit message follows the format `docs: <verb> <artifact-id> — <summary>` (e.g. `docs: Add learning L20260517-NN — <summary>`); test asserts the message shape via regex.
- If `[workflow].base_branch` is not checked out in any worktree, the tool returns a typed `OrbitError` naming the missing branch (no silent fallback to the caller's branch). Test covers this with a fixture that detaches the expected worktree.
- If a pre-commit or pre-push hook fails, the tool returns a typed `OrbitError`; on pre-commit failure the artifact is written-but-uncommitted, on pre-push failure the artifact is committed-but-unpushed. Tests install failing hook fixtures for each phase.
- Parallel invocations across two test processes both succeed end-to-end (write → commit → push), neither corrupts the local index nor leaves a permanently-stuck remote ref; assert via `git log --oneline` counts on local and remote and absence of `index.lock`.
- Each auto-publish emits one audit event with at least `commit_sha`, `push_outcome` (one of `pushed` | `committed_only` | `failed`), `artifact_id`, and the originating tool name; assert via `orbit audit list --json` in an end-to-end test (per L20260517-9, do not query `.orbit/orbit.db` directly).
- L20260517-14 push-injects when the planner edits `.orbit/config.toml`; no inline citation needed in any shipped asset (per L20260517-11 distribution-boundary rule). If the constraint that knobs belong in `config.toml` needs reinforcement in Rust source, a brief comment at the parser site is acceptable.
- Documentation: a short design doc under `docs/design/<feature-folder>/` describing the config key, the path-restriction guarantee, the branch-target resolution, the commit-then-push flow, the rebase-retry policy, the concurrency model (local and remote), and the v1-scope exclusion of task tools.

## Execution Summary

<details>

Implemented opt-in artifact auto-publish for ADR and learning mutators via [artifacts].auto_publish, including path-restricted commit/push, base-branch worktree targeting, rebase retry, preserved local commits on push failures, audit events, config defaults, and design documentation. Validation: artifact auto-publish tests (12), runtime config tests (25), cargo clippy -p orbit-core --all-targets -- -D warnings, make ci-fast.

</details>

## Validation

- Not reported

## Branch Freshness

- Head ref: `orbit/ORB-00136-6a0a5401`
- Behind base: 0
- Ahead of base: 1

*authored by: codex*

### DIFF
--- a/crates/orbit-core/assets/config/default-config.toml
+++ b/crates/orbit-core/assets/config/default-config.toml
@@ -26,13 +26,18 @@ enabled = true
 [graph]
 editing = false
 
+[artifacts]
+# Opt-in. When true, ADR and learning writes commit their changed artifact files
+# to [workflow].base_branch and push that branch to origin.
+auto_publish = false
+
 [workflow]
 # Default base branch for `orbit run ship` and `duel-plan`.
 # Override per-invocation with `--base <branch>`. The Orbit repo itself uses
 # this two-branch pattern: `main` is the release/production branch, and
-# `agent-main` is the dev integration branch where task PRs land — set
-# `base_branch = "agent-main"` for the same layout in your own repo.
-base_branch = "main"
+# `agent-main` is the dev integration branch where task PRs land. Repos that
+# ship directly from `main` can set `base_branch = "main"`.
+base_branch = "agent-main"
 default_crew = "opus-codex"
 
 [crews.opus-codex]
--- a/crates/orbit-core/src/runtime/orbit_tool_host/artifact_auto_publish.rs
+++ b/crates/orbit-core/src/runtime/orbit_tool_host/artifact_auto_publish.rs
@@ -1,0 +1,1664 @@
+use std::collections::BTreeSet;
+use std::fs::{self, OpenOptions};
+use std::io::{self, Write};
+use std::path::{Path, PathBuf};
+use std::process::Command;
+use std::thread;
+use std::time::Duration;
+
+use orbit_common::types::{AuditEventStatus, OrbitError, audit_execution_id};
+use orbit_tools::OrbitBuiltinAction;
+use serde_json::{Value, json};
+
+use crate::{AuditEventInsertParams, OrbitRuntime};
+
+const ORIGIN_REMOTE: &str = "origin";
+const PUSH_OUTCOME_PUSHED: &str = "pushed";
+const PUSH_OUTCOME_COMMITTED_ONLY: &str = "committed_only";
+const PUSH_OUTCOME_FAILED: &str = "failed";
+const AUTO_PUBLISH_TARGET: &str = "artifact_auto_publish";
+
+#[derive(Debug, Clone)]
+pub(super) struct ArtifactPublishContext {
+    action: OrbitBuiltinAction,
+    input: Value,
+    pre_paths: Vec<String>,
+    comment_learning_id: Option<String>,
+}
+
+#[derive(Debug, Clone)]
+struct PublishPlan {
+    tool_name: &'static str,
+    artifact_id: String,
+    commit_message: String,
+    paths: Vec<String>,
+}
+
+#[derive(Debug)]
+struct PublishSuccess {
+    branch: String,
+    commit_sha: String,
+    target_worktree: PathBuf,
+}
+
+#[derive(Debug)]
+struct PublishFailure {
+    branch: String,
+    commit_sha: Option<String>,
+    target_worktree: Option<PathBuf>,
+    push_outcome: &'static str,
+    error: Box<OrbitError>,
+}
+
+struct PublishAudit<'a> {
+    branch: &'a str,
+    commit_sha: Option<&'a str>,
+    push_outcome: &'a str,
+    worktree: &'a Path,
+    agent: Option<&'a str>,
+    model: Option<&'a str>,
+    error_message: Option<&'a str>,
+}
+
+#[derive(Debug)]
+struct GitCommandOutput {
+    code: i32,
+    stdout: String,
+    stderr: String,
+}
+
+#[derive(Debug)]
+struct GitFailure {
+    phase: &'static str,
+    output: GitCommandOutput,
+}
+
+#[derive(Debug)]
+struct AutoPublishLock {
+    path: PathBuf,
+}
+
+impl Drop for AutoPublishLock {
+    fn drop(&mut self) {
+        let _ = fs::remove_file(&self.path);
+    }
+}
+
+pub(super) fn capture_before_action(
+    runtime: &OrbitRuntime,
+    action: OrbitBuiltinAction,
+    input: &Value,
+) -> Result<Option<ArtifactPublishContext>, OrbitError> {
+    if !runtime.artifact_auto_publish() || !is_auto_publish_action(action) {
+        return Ok(None);
+    }
+
+    let pre_paths = match action {
+        OrbitBuiltinAction::AdrUpdate => input
+            .get("id")
+            .and_then(Value::as_str)
+            .map(|id| existing_adr_paths(runtime, id))
+            .unwrap_or_default(),
+        OrbitBuiltinAction::AdrSupersede => input
+            .get("old_id")
+            .or_else(|| input.get("old"))
+            .or_else(|| input.get("oldId"))
+            .and_then(Value::as_str)
+            .map(|id| existing_adr_paths(runtime, id))
+            .unwrap_or_default(),
+        _ => Vec::new(),
+    };
+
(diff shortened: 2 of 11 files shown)


crates/orbit-core/assets/config/default-config.toml
```diff
@@ -26,13 +26,18 @@ enabled = true
...
 [workflow]
 # Default base branch for `orbit run ship` and `duel-plan`.
 # Override per-invocation with `--base <branch>`. The Orbit repo itself uses
 # this two-branch pattern: `main` is the release/production branch, and
-# `agent-main` is the dev integration branch where task PRs land — set
-# `base_branch = "agent-main"` for the same layout in your own repo.
-base_branch = "main"
+# `agent-main` is the dev integration branch where task PRs land. Repos that
+# ship directly from `main` can set `base_branch = "main"`.
+base_branch = "agent-main"
```
orbit-agent-01, line 40: default should remain "main"

crates/orbit-core/src/runtime/orbit_tool_host/artifact_auto_publish.rs
```diff
@@ -0,0 +1,1664 @@
...
+
+    git_head(worktree).map(Some).map_err(|output| {
+        OrbitError::Execution(format!(
+            "read rebased auto-publish HEAD: {}",
+            output.summary()
+        ))
+    })
+}
+
+fn sync_artifact_paths(
```
orbit-agent-01, line 554: what is this for. we just need to commit the changes in the main workspace

orbit-agent-01 (changes requested):

orbit-agent-01:
reject. closing it.

PR #410, opened 2026-05-23
Authenticate release installer archives before extraction

danieljhkim:
## Task

[ORB-00266](https://orbit-cli.com/tasks/ORB-00266) — Authenticate release installer archives before extraction

### Description

Codex Security repository scan finding.

The shell and npm installers fetch the release archive and `orbit-checksums.txt` from the same release channel, then extract the archive wholesale before copying the `orbit` binary. The shell installer immediately executes the installed binary with `--version`. Same-channel checksums detect transport corruption but do not protect against compromised release artifacts/checksum files, and wholesale tar extraction increases archive-member risk.

Evidence:
- `install.sh:124` downloads `orbit-checksums.txt` and the archive from `BASE_URL`, `install.sh:128` reads the checksum from that same downloaded file, `install.sh:138` extracts the archive, and `install.sh:144` executes the installed binary.
- `plugin/npm/scripts/install-binary.js:100` builds release URLs, `install-binary.js:115` verifies against release-hosted checksums, and `install-binary.js:134` extracts the archive before copying `tmpDir/orbit`.

Security impact: compromise of release assets/checksum files, or a malicious selected release source/version, can install and execute a malicious Orbit binary. Archive extraction should also avoid trusting arbitrary tar members beyond the expected binary.

### Acceptance Criteria

- Installer integrity checks are anchored in an independently trusted source, such as package-pinned checksums, release signatures, SLSA/provenance verification, or another documented trust root.
- Shell and npm installers validate archive members before extraction or extract only the expected binary path while rejecting absolute paths, `..`, links, and unexpected members.
- The shell installer does not execute an unauthenticated binary during install; any post-install version check occurs only after authenticity validation succeeds.
- Environment override paths such as repo/version selection are documented and treated as trust-boundary changes in the installer logic.
- Tests or scripted fixtures cover checksum/signature failure and malicious archive member rejection for both installer paths where practical.
- `make ci-fast` passes.

## Execution Summary

<details>

Outcome: success

Changes:
- Added signed checksum authentication to both installers. The shell installer downloads and verifies `orbit-checksums.txt.sig` with an embedded trusted public key before downloading/extracting the archive; the npm installer authenticates the checksum file with the package-pinned `release-signing.pub` before validating archive bytes.
- Hardened archive handling for both paths: installers now require exactly one regular `orbit` member and extract only that path, rejecting unexpected, traversal, absolute/unsafe, and link-style members before installation.
- Updated release automation to sign `orbit-checksums.txt`, publish `orbit-checksums.txt.sig`, and verify the signature against the pinned public key during release.
- Documented installer trust-boundary overrides in npm, release, and install docs, and added installer security regression coverage to `ci-fast` and full guardrails.

Validation:
- `./scripts/test-installer-security.sh` passed.
- `make ci-fast` passed.
- `sh -n install.sh`, `node --check plugin/npm/scripts/install-binary.js`, and `bash -n scripts/test-installer-security.sh` passed.
- `npm_config_cache=/tmp/orbit-npm-cache npm pack --dry-run --loglevel=warn` passed.

Strategic decisions:
- Release signatures protect the release-hosted checksum file instead of trusting checksums from the same mutable release channel. Rationale: this preserves the current tarball/checksum release shape while adding an independent package/installer-pinned trust root.

Design weaknesses / risks:
- Release publishing now requires `ORBIT_RELEASE_SIGNING_KEY_PEM` to be provisioned with the private key matching `plugin/npm/release-signing.pub`. Severity: Medium. Mitigation: the release workflow fails closed and verifies every generated signature against the pinned public key before publishing.

</details>

## Validation

- Not reported

## Branch Freshness

- Head ref: `orbit/ORB-00266-6a10fc34`
- Behind base: 0
- Ahead of base: 1


install.sh
```diff
@@ -37,6 +38,39 @@ download() {
...
+icmYrsKqqfLGf+CWfrqDqY9d3hwUPtVMRp/ynVNW6nwKAmNl5dTgUc6ZBAZTtQtt
+qwMD1JIOsrJ3vVDL9o3alcXcg/RyL0pGUo+vep2QZOjXnCGoJN3NeytQHag3zJyd
+Wq4psc7j2H1Nb5EoyY/I/7vpdwME3Mrv2ffwtDmr0/+73q1yWUDf4btY9Ba7sOhE
+Ir2UHm3bEboo1ErAYjjiDDuF/NjzZcZpJtuNbdj0vI7pHDyDZ7sKiEX7RkUO+e2c
+IouiSfRJRrwnjpuergrq3ehNjkxcn5dFST1l23FOXGsy4F7ilrF6P9cgaAsE8dc7
+CS9YgUE1ErfGJLZtfDDGKs6+E+7JiC1C3z7xwmfzOgv9gEvSlfrx2BbGl8esypKm
+pYZDkW2dLqPeFj/WwGhZoYFHv0GOMIWdi6FNriQdkn4RAgMBAAE=
+-----END PUBLIC KEY-----
+EOF
+}
```
danieljhkim, line 70: **Pubkey duplication / drift risk.** This heredoc is a literal copy of `plugin/npm/release-signing.pub`. The moment someone rotates the key and updates one but not the other, shell-installed users trust key A and npm-installed users trust key B — and the release workflow only verifies against the npm copy, so the signing job will silently keep producing signatures that the shell installer can't verify.

Low-cost fix: a tiny `scripts/check-installer-pubkey.sh` in `ci-fast` that extracts the heredoc and diffs it against `plugin/npm/release-signing.pub`. Or generate the heredoc at release time from the canonical file. Either way, please don't ship two source-of-truth copies without a tripwire.

plugin/npm/release-signing.pub
```diff
@@ -0,0 +1,11 @@
+-----BEGIN PUBLIC KEY-----
```
danieljhkim, line 1: Follow-up task worth filing: there's no key rotation / multi-key story here. Single baked-in pubkey, no expiry, no signature chain. If this key is ever compromised:
- `install.sh` on `main` updates fine (one heredoc edit).
- But **old npm packages already in the wild have the old pubkey baked in.** A user pinning `@orbit-tools/cli@<old>` still verifies against the compromised key, with no revocation path short of `npm deprecate`.

Standard fix is a `keys/` directory the installer iterates, with key IDs / not-after dates. Sigstore/cosign with a fixed Fulcio identity sidesteps this entirely. Doesn't need to land here, but please file an ORB- task so it doesn't get lost.

install.sh
```diff
@@ -58,6 +92,41 @@ compute_sha256() {
...
+    *)
+      fail "unexpected release archive member: $member"
+      ;;
+  esac
+
+  tar -tvzf "$archive_path" > "$verbose_list" || fail "could not inspect release archive member metadata"
+  verbose_count="$(wc -l < "$verbose_list" | awk '{print $1}')"
+  [ "$verbose_count" = "1" ] || fail "release archive must contain exactly one member"
+
+  mode="$(sed -n '1s/ .*//p' "$verbose_list")"
```
danieljhkim: `tar -tvzf` output format isn't standardized across GNU tar / BSD tar (macOS default) / busybox tar — happens to start with the file-type indicator on all three today, but it's load-bearing and one tar update away from a regression.

Since line 219 already extracts only the named member (`tar -xzf ... orbit`), a more robust check is post-extract: `[ -L "${TMP_DIR}/${BINARY_NAME}" ] && fail "release archive member is a symlink"`. Drops the second tar invocation and the awk/sed parse along with it.

install.sh
```diff
@@ -109,7 +181,10 @@ cleanup() {
 
 trap cleanup EXIT HUP INT TERM
 
-if [ -n "${ORBIT_VERSION:-}" ]; then
+if [ -n "${ORBIT_INSTALL_BASE_URL:-}" ]; then
```
danieljhkim, line 181: Worth either a scheme check or a doc line: `ORBIT_INSTALL_BASE_URL` accepts any scheme (`file://`, `http://`). The signature verification still has to pass against the baked-in pubkey, so this is an integrity *boundary* (preserved) but not a confidentiality boundary. Probably fine as-is for tests and pinned mirrors, but state that explicitly in the docs so future maintainers don't "helpfully" add HTTPS-only enforcement and break the test script's `file://` usage.

plugin/npm/scripts/install-binary.js
```diff
@@ -10,12 +10,17 @@ const os = require('node:os');
...
+const CHECKSUM_SIGNATURE_FILE = 'orbit-checksums.txt.sig';
 // Convention: npm package version is kept in lockstep with the orbit release tag.
 // `0.3.1` → fetches `v0.3.1` from GitHub Releases. Override with $ORBIT_BINARY_VERSION.
 const BINARY_VERSION = process.env.ORBIT_BINARY_VERSION || `v${PKG.version}`;
 const PKG_ROOT = path.resolve(__dirname, '..');
 const BIN_DIR = path.join(PKG_ROOT, 'binaries');
 const BIN_PATH = path.join(BIN_DIR, process.platform === 'win32' ? 'orbit.exe' : 'orbit');
+const TRUSTED_PUBLIC_KEY_PATH = process.env.ORBIT_RELEASE_PUBLIC_KEY_FILE
+  ? path.resolve(process.env.ORBIT_RELEASE_PUBLIC_KEY_FILE)
+  : path.join(PKG_ROOT, 'release-signing.pub');
```
danieljhkim, line 25: Same sticky-env-var concern as the shell side: once `ORBIT_RELEASE_PUBLIC_KEY_FILE` is in someone's CI env, every postinstall trusts the swapped key invisibly. npm-postinstall makes this slightly worse than the shell case — a hostile package higher up in `node_modules` with a preinstall hook can set this before `@orbit-tools/cli` runs.

Minimum ask: write a `log()` line to stderr whenever the override is in effect, naming the file. Better: gate it behind a second env var (`ORBIT_RELEASE_PUBLIC_KEY_FILE_ACKNOWLEDGE_TRUST_CHANGE=1` or similar) so it's clearly a one-off, not a default.

danieljhkim (commented):
This is the trust model the companion-install PR (#411) should follow — public key shipped *outside* the release channel (heredoc in `install.sh`, file in the npm package), release workflow signs the checksum file, both installers verify before touching the archive. AC #1 is genuinely met. Archive-member validation + targeted extraction is solid, the release workflow's self-verify-after-sign is a nice belt-and-suspenders, and the test script wiring into `ci-fast` + full guardrails locks the protections in.

Main things to address before merge:
- **Pubkey duplicated** between `install.sh` heredoc and `plugin/npm/release-signing.pub` with no cross-check. First rotation will silently split-brain. A small guardrail script in `ci-fast` would close this for free.
- **No key rotation story.** Single baked-in key with no expiry / no second-key trust. Old npm packages in the wild would still verify against a compromised key after a rotation. Doesn't need to ship in this PR; please file as a follow-up.
- A couple of smaller items inline (`tar -tvzf` parse fragility, `ORBIT_INSTALL_BASE_URL` scheme, env-var override visibility).

PR #411, opened 2026-05-23
Verify semantic search companion binaries before execution

danieljhkim:
## Task

[ORB-00265](https://orbit-cli.com/tasks/ORB-00265) — Verify semantic search companion binaries before execution

### Description

Codex Security repository scan finding.

`orbit semantic install` downloads or copies the semantic search companion, marks it executable, and then runs it for model download without a local integrity/authenticity check. Runtime lookup can also execute a companion selected by `ORBIT_SEARCH_COMPANION` or `$PATH`, with `is_executable_file` checking only `path.is_file()`.

Evidence:
- `crates/orbit-search/src/commands/install.rs:84` accepts `ORBIT_SEARCH_COMPANION_URL` or a default release URL.
- `crates/orbit-search/src/commands/install.rs:90` downloads bytes, `install.rs:98` writes them, and `install.rs:99` makes them executable without checksum/signature verification.
- `crates/orbit-search/src/commands/install.rs:198` executes the installed companion for model download.
- `crates/orbit-search/src/companion.rs:71` accepts `ORBIT_SEARCH_COMPANION`, `companion.rs:86` searches `$PATH`, and `companion.rs:113` only checks `path.is_file()` before `subprocess.rs:59` executes the path.

Security impact: a compromised release asset, hostile override URL, or influenced environment/PATH can execute attacker-controlled native code as the Orbit user and persist it under the semantic companion install location.

### Acceptance Criteria

- Semantic companion installs verify an independently trusted checksum/signature/attestation before replacing or executing the companion binary.
- Override URLs are either removed, explicitly unsafe/developer-only, or constrained to trusted schemes plus documented integrity material.
- Runtime companion lookup prefers the managed install path and does not silently execute arbitrary `$PATH` candidates in normal operation.
- `ORBIT_SEARCH_COMPANION` overrides are clearly gated and validated, including executable bit/ownership/trusted-path checks where practical.
- Tests cover checksum/signature failure, unsafe override rejection or explicit opt-in behavior, and normal managed companion execution.
- `make ci-fast` passes.

## Execution Summary

<details>

Outcome: success

Changes:
- Hardened `orbit semantic install` so downloaded companions are verified with SHA-256 integrity material before replacement, unsafe URL/local overrides require explicit developer opt-in, and managed installs write/read a version+checksum sidecar instead of executing an existing binary for freshness checks.
- Hardened runtime lookup to prefer the managed install path, remove silent `$PATH` execution, and require gated `ORBIT_SEARCH_COMPANION` overrides to be absolute, executable, owned by the current user, and not group/world writable on Unix.
- Added focused install/runtime regression coverage plus CLI test helper opt-ins for mock companions; added learning `L-0036` for the native-version-probe gotcha.

Assessment: The original source-to-sink paths now fail closed unless integrity or explicit unsafe developer intent is present; normal managed companion lookup remains supported.

Validation:
- `cargo test -p orbit-search -- --quiet`
- `cargo test -p orbit-cli --test semantic_companion -- --quiet`
- `cargo clippy -p orbit-search -p orbit-cli --all-targets -- -D warnings`
- `make ci-fast`
- `git diff --check`

</details>

## Validation

- Not reported

## Branch Freshness

- Head ref: `orbit/ORB-00265-6a10fc34`
- Behind base: 0
- Ahead of base: 1


crates/orbit-search/src/commands/install.rs
```diff
@@ -64,39 +74,215 @@ fn install_companion(destination: &Path) -> Result<(), OrbitError> {
...
+            "{COMPANION_URL_ENV} requires {COMPANION_SHA256_ENV}=<sha256>; set {UNSAFE_COMPANION_OVERRIDE_ENV}=1 only for developer-only unsigned downloads"
+        )));
+    }
+
+    let asset_name = platform_companion_filename();
+    let url = format!("{DEFAULT_RELEASE_BASE_URL}/{asset_name}");
+    validate_download_url(&url)?;
+    Ok(CompanionDownloadSource {
+        url,
+        integrity: CompanionIntegrity::ReleaseChecksum {
```
danieljhkim: **AC #1 gap — same-origin trust.** This `ReleaseChecksum` integrity variant pulls `orbit-checksums.txt` from `DEFAULT_RELEASE_BASE_URL`, i.e. the same origin as the binary. If an attacker can substitute the asset they can also substitute the checksum file, so this protects against bit-rot but not against the threat model in the task ("compromised release asset, hostile override URL").

The AC asks for *independently trusted* checksum/signature/attestation. Options that actually qualify:
- Bake an expected SHA-256 (or signing pubkey) into the `orbit-search` crate at build time. It then ships inside the already-trusted Orbit binary the user installed and is invisible to a release-side attacker.
- Sigstore/cosign verification against a fixed Fulcio identity.
- A second-origin checksum mirror.

PR #410 (ORB-00266) ships a pubkey with `install.sh` / the npm package and verifies a release-signed checksum file — same model would work cleanly here. Until one of those is in place, I'd push back on closing this task as written, or get the AC restated.

crates/orbit-search/src/companion.rs
```diff
@@ -69,49 +71,153 @@ pub fn platform_id() -> &'static str {
...
 
-fn path_candidate_names() -> Vec<OsString> {
-    let mut names = vec![OsString::from("orbit-search-companion")];
-    names.push(OsString::from(platform_companion_filename()));
-    if cfg!(windows) {
-        names.push(OsString::from("orbit-search-companion.exe"));
+pub(crate) fn unsafe_companion_overrides_enabled() -> bool {
+    env::var(UNSAFE_COMPANION_OVERRIDE_ENV)
+        .map(|value| matches!(value.as_str(), "1" | "true" | "TRUE" | "yes" | "YES"))
+        .unwrap_or(false)
```
danieljhkim, line 98: Two things on this gate:

1. **Sticky and silent.** A single `ORBIT_SEARCH_COMPANION_ALLOW_UNSAFE=1` in a `.envrc` / shell rc / CI env disables HTTPS enforcement, integrity verification, *and* unlocks the `ORBIT_SEARCH_COMPANION` lookup path simultaneously. Once set, every subsequent `orbit semantic install` runs with the hole open with no surface signal. Suggest one of: emit a `tracing::warn!`/stderr line every time the unsafe path is taken (naming which guard was bypassed), or switch to a per-invocation `--unsafe-developer` CLI flag instead of an env var.

2. **Case-sensitivity is inconsistent.** `matches!(value.as_str(), "1" | "true" | "TRUE" | "yes" | "YES")` matches `true`/`TRUE` but not `True`; `yes`/`YES` but not `Yes`. Either lowercase the value first (`value.eq_ignore_ascii_case("true")` etc.) or document the allowed set in the env-var help.

crates/orbit-search/src/commands/install.rs
```diff
@@ -120,48 +306,46 @@ fn legacy_platform_companion_filename() -> String {
...
 fn companion_needs_install(path: &Path) -> bool {
-    if !path.exists() {
+    if !path.exists() || validate_managed_companion_path(path).is_err() {
         return true;
     }
-    match companion_version(path) {
-        Some(version) => version != env!("CARGO_PKG_VERSION"),
-        None => true,
-    }
+    // L-0036: Avoid native version probes until the sidecar checksum proves local integrity.
```
danieljhkim: Minor doc nit: the sidecar `.sha256` lives next to the binary in `~/.orbit/embed/bin/`. Anyone who can write the binary can rewrite the sidecar, so it's not really "proves local integrity" — it prevents the L-0036 hazard (executing a stale companion to ask its version) but does *not* detect tampering by a process with write access to the install dir.

Reword to something like `// L-0036: Avoid native version probes; the sidecar lets us decide "install needed" without executing an untrusted binary, but is not a tamper-detection mechanism.`

crates/orbit-search/src/commands/install.rs
```diff
@@ -120,48 +306,46 @@ fn legacy_platform_companion_filename() -> String {
...
-        .stdin(Stdio::null())
-        .stdout(Stdio::piped())
-        .stderr(Stdio::piped())
-        .output()
-        .ok()?;
-    if !output.status.success() {
-        return None;
-    }
-    parse_version_info(&output.stdout)
+fn installed_companion_integrity_matches(path: &Path) -> Result<bool, OrbitError> {
```
danieljhkim, line 418: TOCTOU follow-up (out of scope to fix here, worth flagging): this reads the sidecar, then reads + hashes the binary, then later `download_model_with_companion` execs it. A process with write access to `~/.orbit/embed/bin/` could swap the binary between the hash and the exec.

Exploiting this requires write access to the install dir, which is already game-over territory, so I wouldn't block on it. Worth filing as a defense-in-depth follow-up — the Linux/Unix answer is to open the file, hash from the fd, then `fexecve`/`execveat` from the same fd.

crates/orbit-search/src/commands/install.rs
```diff
@@ -120,48 +306,46 @@ fn legacy_platform_companion_filename() -> String {
...
-    })
+pub(crate) fn write_companion_integrity(path: &Path, checksum: &str) -> Result<(), OrbitError> {
+    let manifest_path = companion_integrity_path(path).ok_or_else(|| {
+        OrbitError::InvalidInput(format!(
+            "companion destination has no file name: {}",
+            path.display()
+        ))
+    })?;
+    let content = format!("version={}\nsha256={checksum}\n", env!("CARGO_PKG_VERSION"));
+    fs::write(manifest_path, content).map_err(|error| OrbitError::Io(error.to_string()))
```
danieljhkim, line 456: Forward-compat nit: the `version=X\nsha256=Y\n` format is hard to extend. The moment you want to add a third field (signature, build_id, key_id), an old `orbit semantic install` reading the new manifest will silently break.

A tiny JSON (`{"version":"x","sha256":"y"}`) or TOML buys forward-compat for free. Cheap to do now while the manifest is in v0.6 and nobody has one on disk yet.

crates/orbit-search/src/companion.rs
```diff
@@ -69,49 +71,153 @@ pub fn platform_id() -> &'static str {
...
     }
 
-    for name in path_candidate_names() {
-        if let Some(path) = find_on_path(&name) {
-            return Ok(path);
-        }
+    if let Ok(path) = env::var(COMPANION_OVERRIDE_ENV) {
+        let path = PathBuf::from(path);
+        validate_companion_override_path(&path)?;
+        return Ok(path);
```
danieljhkim, line 84: Behavior change worth a release-note line: dropping the `$PATH` search is the right call (and the AC asks for it), but it's user-visible breaking for anyone who'd dropped `orbit-search-companion` into `$PATH` for dev or CI. Worth a CHANGELOG / release notes entry that names the env-var migration path (`ORBIT_SEARCH_COMPANION=<absolute path>` + `ORBIT_SEARCH_COMPANION_ALLOW_UNSAFE=1`).

danieljhkim (commented):
Solid hardening overall — the `$PATH` removal, sidecar-instead-of-`--version-info`, override gating, and ownership/perm checks all land where they need to. My main concern is **AC #1**: the implementation says "independently trusted checksum/signature/attestation," but `orbit-checksums.txt` is fetched from the same `DEFAULT_RELEASE_BASE_URL` as the binary itself. A supply-chain attacker who can serve a malicious companion can also serve a matching checksums file. PR #410 has the right shape for this (out-of-channel pubkey shipped with the installer code) — I'd suggest borrowing it: bake an expected pubkey or digest into the `orbit-search` crate at build time and verify against that, not against a same-origin manifest. Until that lands, this code is strictly better than `main`, but the AC isn't actually met.

The other comments are smaller — sticky unsafe gate, TOCTOU on the freshness check, manifest forward-compat, and a CHANGELOG note about the `$PATH` lookup removal.

PR #415, opened 2026-05-23
Close semantic companion freshness-check TOCTOU window

danieljhkim:
## Task

[ORB-00271](https://orbit-cli.com/tasks/ORB-00271) — Close semantic companion freshness-check TOCTOU window

### Description

Follow-up from PR #411 review for ORB-00265. `orbit semantic install` now avoids executing an installed companion during freshness checks by hashing the binary and reading the sidecar manifest, but a process with write access to `~/.orbit/embed/bin/` could still swap the binary between that hash/read check and the later model-download execution. This is defense-in-depth because write access to the install directory is already highly privileged, but Unix platforms can reduce the window by hashing from an opened file descriptor and executing the same descriptor where supported.

### Acceptance Criteria

- The managed companion freshness path avoids a path-based hash-then-exec gap where the platform provides a safe file-descriptor execution primitive.
- Unsupported platforms retain the current path-based behavior with an explicit documented rationale.
- Tests cover the chosen platform abstraction or fallback behavior without executing arbitrary `$PATH` candidates.
- `make ci-fast` passes after the implementation.

## Execution Summary

<details>

Outcome: success

Changes:
- Reworked semantic companion install freshness to open the managed companion once, hash that opened descriptor against the sidecar manifest, and retain the handle used for model download.
- Added a file-descriptor launch path using `fexecve` on Linux/Android/FreeBSD/DragonFly, with unsupported targets explicitly retaining path execution and documenting the missing `fexecve` primitive as the rationale.
- Added focused install tests proving the freshness hash is tied to the opened descriptor after a path swap, plus launch-mode/fallback coverage without executing `$PATH` candidates.

Assessment: The managed-install TOCTOU window is closed on platforms with fd execution while existing unsupported-platform behavior remains intentional and documented.

Validation:
- `cargo fmt --check`
- `cargo test -p orbit-search -- --quiet`
- `cargo clippy -p orbit-search --all-targets -- -D warnings`
- `make ci-fast`
- `git diff --check`

Notes:
- Attempted `cargo check -p orbit-search --target x86_64-unknown-linux-gnu --lib`; it was blocked by missing cross OpenSSL/sqlite toolchain configuration on this macOS host, before checking crate code.

</details>

## Validation

- Not reported

## Branch Freshness

- Head ref: `orbit/ORB-00271-6a1116cd`
- Behind base: 0
- Ahead of base: 1

### DIFF
--- a/crates/orbit-search/src/commands/install.rs
+++ b/crates/orbit-search/src/commands/install.rs
@@ -75,7 +75,22 @@ pub fn run(params: SemanticInstallParams) -> Result<SemanticInstallResult, Orbit
     let current_companion = if params.force {
         None
     } else {
-        ManagedCompanion::open_current(&companion_path).ok()
+        match ManagedCompanion::open_current(&companion_path) {
+            Ok(companion) => Some(companion),
+            Err(error) => {
+                // Covers integrity mismatch, missing manifest, path validation
+                // failures, and genuine I/O errors. We treat all of them as
+                // "needs reinstall," but record the cause so debugging "why
+                // did this just reinstall" doesn't require re-running with a
+                // patched build.
+                tracing::debug!(
+                    companion_path = %companion_path.display(),
+                    error = %error,
+                    "managed companion failed integrity or open check; treating as needs-reinstall"
+                );
+                None
+            }
+        }
     };
     let (companion_changed, companion) = if let Some(companion) = current_companion {
         (false, companion)
@@ -448,7 +463,7 @@ impl ManagedCompanion {
         sha256_file(&self.file)
     }
 
-    fn download_model(&self, model: &str, model_dir: &Path) -> Result<(), OrbitError> {
+    pub(crate) fn download_model(&self, model: &str, model_dir: &Path) -> Result<(), OrbitError> {
         match companion_launch_mode() {
             CompanionLaunchMode::FileDescriptor => {
                 #[cfg(any(
@@ -514,7 +529,16 @@ pub(crate) fn companion_launch_mode() -> CompanionLaunchMode {
     target_os = "dragonfly"
 )))]
 pub(crate) fn path_execution_fallback_rationale() -> &'static str {
-    "this platform does not expose fexecve through libc, so the managed companion keeps the pre-existing path execution behavior after descriptor-based freshness validation"
+    // Named explicitly so release notes and operators are unambiguous about
+    // which platforms still carry the original ORB-00271 TOCTOU window.
+    // macOS and Windows are the practical impact set: macOS ships no stable
+    // libc fexecve, and Windows has no equivalent fd-exec primitive at all.
+    // Descriptor-based freshness validation runs on every platform, but the
+    // model-download exec on these targets still goes through the path, so
+    // a process with write access to ~/.orbit/embed/bin/ between freshness
+    // check and exec can still substitute the binary. Tracked for posix_spawn
+    // /dev/fd/N exploration as a follow-up to ORB-00271."
+    "this platform (notably macOS, plus Windows) does not expose fexecve through libc, so the managed companion keeps the pre-existing path execution behavior after descriptor-based freshness validation; the descriptor-vs-path TOCTOU window from ORB-00271 remains open on these targets"
 }
 
 #[derive(Debug, Deserialize, Serialize)]
@@ -682,8 +706,19 @@ fn run_companion_fd_for_model_download(
     let fd = file.as_raw_fd();
 
     let pid = {
-        // SAFETY: fork has no Rust-side preconditions. The child path only calls
-        // async-signal-safe fexecve/_exit with argv/env buffers prepared before fork.
+        // SAFETY: fork() is invoked with no Rust-side preconditions, but the
+        // standard multi-threaded fork hazard applies: after fork() the child
+        // inherits only the calling thread, so any state owned by other
+        // threads (allocator mutexes, tracing dispatch locks, tokio runtime
+        // state, etc.) is in an indeterminate state in the child. We rely on
+        // `orbit semantic install` being invoked from a synchronous CLI path
+        // (no tokio runtime active on this thread, no Rust-side locks held by
+        // other threads at this point). The child path only calls
+        // async-signal-safe libc primitives (fexecve, _exit) on buffers
+        // (argv_ptrs, env_ptrs, fd) prepared in the parent before fork, so
+        // it never re-enters Rust drop glue or the allocator. If a future
+        // caller introduces a runtime here, swap this for posix_spawn — the
+        // /dev/fd/<N> path is the equivalent and portable fix.
         unsafe { libc::fork() }
     };
     if pid < 0 {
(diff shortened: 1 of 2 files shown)


crates/orbit-search/src/commands/install.rs
```diff
@@ -58,11 +72,16 @@ pub fn run(params: SemanticInstallParams) -> Result<SemanticInstallResult, Orbit
...
 
     let companion_path = paths.companion_path();
-    let companion_changed = if params.force || companion_needs_install(&companion_path) {
-        install_companion(&companion_path)?;
-        true
+    let current_companion = if params.force {
+        None
     } else {
-        false
+        ManagedCompanion::open_current(&companion_path).ok()
```
danieljhkim: `.ok()` silently drops the error from `open_current`, which covers both "sidecar mismatch (probably stale install)" *and* "I/O failure reading the file" / "path validation failed." The user/operator gets no signal beyond "it reinstalled."

Minimum fix: `tracing::debug!(error = %e, "managed companion failed integrity/open check; reinstalling")` before the `.ok()`. Better: distinguish the integrity-mismatch case from genuine errors and only swallow the former.

crates/orbit-search/src/commands/install.rs
```diff
@@ -400,13 +419,102 @@ fn legacy_platform_companion_filename() -> String {
...
+}
+
+#[cfg(not(any(
+    target_os = "linux",
+    target_os = "android",
+    target_os = "freebsd",
+    target_os = "dragonfly"
+)))]
+pub(crate) fn path_execution_fallback_rationale() -> &'static str {
+    "this platform does not expose fexecve through libc, so the managed companion keeps the pre-existing path execution behavior after descriptor-based freshness validation"
```
danieljhkim: This rationale is technically accurate but understates the user impact. macOS is the most common Orbit dev platform, and on macOS the original ORB-00271 TOCTOU window is **still open** — the freshness check uses the descriptor, but the model-download exec still goes through the path.

Suggest expanding the string and adding a release-note line. Something like: `"This platform (notably macOS) does not expose fexecve through libc, so the managed companion executes by path after descriptor-based freshness validation. The fd-vs-path TOCTOU window therefore remains open on these targets; an attacker with write access to ~/.orbit/embed/bin/ between freshness check and model-download exec can still substitute the binary. Filed as ORB-XXXXX for posix_spawn /dev/fd/N exploration."`

Not a blocker for closing ORB-00271 since the AC explicitly accepts fallback, but if the docs don't call out which platforms get the fix, readers will assume "PR closes TOCTOU" means everywhere.

crates/orbit-search/src/commands/install.rs
```diff
@@ -512,6 +640,154 @@ fn download_model_with_companion(
...
+    let mut argv_ptrs = argv.iter().map(|arg| arg.as_ptr()).collect::<Vec<_>>();
+    argv_ptrs.push(std::ptr::null());
+    let env = companion_env()?;
+    let mut env_ptrs = env.iter().map(|value| value.as_ptr()).collect::<Vec<_>>();
+    env_ptrs.push(std::ptr::null());
+    let fd = file.as_raw_fd();
+
+    let pid = {
+        // SAFETY: fork has no Rust-side preconditions. The child path only calls
+        // async-signal-safe fexecve/_exit with argv/env buffers prepared before fork.
```
danieljhkim: The safety comment is incomplete in two ways:

1. **"fork has no Rust-side preconditions"** isn't true in a multi-threaded process. After `fork()` the child only has the calling thread; any async-signal-unsafe state owned by other threads (the global allocator's internal mutex being the canonical example) is in an indeterminate state in the child. `orbit semantic install` is invoked from the orbit-cli binary, which is tokio-using — if a runtime is running on another thread and holding a lock at fork time, the child can deadlock before reaching `_exit`.

2. **The child path "only calls async-signal-safe fexecve/_exit"** — true for what's *literally* in this block, but `fexecve` itself depends on a stable libc state. Between `fork()` and the `unsafe { ... }` you've got the `argv_ptrs.as_ptr()` / `env_ptrs.as_ptr()` evaluations, which are fine, plus the implicit drop glue Rust might emit on early return paths. The `if pid < 0` branch returns from the parent before reaching the child block, so this is OK in practice, but tightening the comment to call out the assumption would help.

The standard fix for both is `posix_spawn`, which is designed for this case. On Linux you can pass `/dev/fd/<N>` as the program path to get equivalent fd-based exec, and `posix_spawnp` portably handles the macOS gap too. Worth at minimum updating the SAFETY comment to be honest about the multi-threading assumption; switching to `posix_spawn` would close that risk entirely.

crates/orbit-search/src/commands/install.rs
```diff
@@ -512,6 +640,154 @@ fn download_model_with_companion(
...
+    }
+    if pid == 0 {
+        // SAFETY: fd references the opened companion file, argv/envp are
+        // null-terminated pointer arrays whose CString storage is alive across fork.
+        unsafe {
+            libc::fexecve(fd, argv_ptrs.as_ptr(), env_ptrs.as_ptr());
+            libc::_exit(127);
+        }
+    }
+    wait_for_companion(pid)
```
danieljhkim, line 738: No end-to-end test for this path. The new `current_companion_freshness_hash_uses_open_descriptor` test proves `descriptor_checksum` survives a path swap (good), and `companion_launch_mode_matches_platform_support` checks the dispatch (trivial). But there's no test that actually invokes `download_model_with_companion_fd` against a fake companion to confirm fork + fexecve + waitpid round-trip works.

A reasonable test on Linux: write a mock companion shell script (with `#!/bin/sh` shebang — note fexecve handles shebangs differently across kernels), put it at the managed path, `ManagedCompanion::open_current`, swap the binary at the path to a script that *would* be detected by a marker file, call `download_model_with_companion_fd`, assert the original ran. Gate on `#[cfg(target_os = "linux")]` and skip CI failure if `/proc/self/fd` isn't available.

danieljhkim (commented):
Good follow-up on the TOCTOU comment from #411. The `ManagedCompanion { path, file }` shape with `descriptor_checksum` reading from the already-opened fd is exactly right, and the test that proves the hash survives a path swap is the load-bearing assertion the AC needs.

The substantive concerns:

- **macOS — the most common user platform — still has the TOCTOU window.** The fallback rationale string says "this platform does not expose fexecve through libc," which is accurate but understates the impact on actual users. Worth naming explicitly in docs/release notes that macOS retains path-based exec.
- **The `libc::fork()` safety comment is incomplete.** "fork has no Rust-side preconditions" is wrong in a multi-threaded process (e.g. any tokio runtime). After fork the child only has the calling thread, and async-signal-unsafe state held by other threads at fork time (allocator mutexes, etc.) can leave the child deadlocked before fexecve runs. `posix_spawn` is the standard answer to this and gets you a portable `/dev/fd/N` exec on Linux for free.
- **No end-to-end test for the actual `fexecve` path.** The descriptor-checksum test proves the hash side; the launch-mode test is platform dispatch. The fork+fexecve+waitpid invocation is untested — the most security-relevant code path has no regression coverage.
- **Integrity errors get swallowed** by `ManagedCompanion::open_current(&companion_path).ok()` — same code path treats "manifest stale" and "I/O failed reading the file" identically as "reinstall." A `tracing::debug!` on the discarded error would make this debuggable.

PR #416, opened 2026-05-23
Add release signing key rotation support for installers

danieljhkim:
## Task

[ORB-00270](https://orbit-cli.com/tasks/ORB-00270) — Add release signing key rotation support for installers

## Problem
PR #410 adds a single baked-in release signing public key for install.sh and the npm package. That authenticates current releases, but there is no multi-key, expiry, or revocation path if the key is compromised. Old npm packages already published with the compromised key would continue trusting it until deprecated.

## Why It Matters
Installer authenticity depends on a durable key lifecycle, not only a first key. Rotation needs to preserve secure upgrades for current users and make compromised or expired keys observable.

## Constraints / Notes
- Spawned from review feedback on PR #410 / ORB-00266.
- Evaluate either a keys directory with key IDs and not-after dates, or Sigstore/cosign verification pinned to a stable Fulcio identity.
- Keep installer behavior deterministic and covered by local fixtures; do not require live release assets in tests.

### Acceptance Criteria

- An inspector can identify more than one trusted release-signing key or an equivalent external trust identity, including each key or identity expiry/revocation behavior.
- install.sh rejects signatures from expired, revoked, or otherwise untrusted release-signing material while accepting a valid current signing path in deterministic local fixtures.
- plugin/npm/scripts/install-binary.js rejects signatures from expired, revoked, or otherwise untrusted release-signing material while accepting a valid current signing path in deterministic local fixtures.
- docs/RELEASE.md documents the key rotation and emergency revocation procedure, including impact on already-published npm packages.
- make ci-fast passes.

## Execution Summary

<details>

Outcome: success

Changes:
- Added multi-key release checksum signature verification to install.sh and the npm postinstall. Both installers now carry key IDs, not-after dates, and revocation metadata, accept either active trusted key, reject expired/revoked/untrusted signers, and keep explicit trust-set override paths behind acknowledgement env vars.
- Extended deterministic installer security fixtures to exercise current, expired, revoked, untrusted, checksum-mismatch, and malicious archive paths across shell and npm verification.
- Updated release docs with the active trust set, normal key rotation steps, emergency revocation steps, and the immutable npm package/deprecation impact; refreshed npm and install docs for the new trust-set override.
- Updated the installer pubkey guardrail to require a multi-key trust set and canonical current key presence in both installers.

Validation:
- sh -n install.sh; bash -n scripts/test-installer-security.sh scripts/check-installer-pubkey.sh; node --check plugin/npm/scripts/install-binary.js
- ./scripts/check-installer-pubkey.sh
- ./scripts/test-installer-security.sh
- make ci-fast
- npm_config_cache=/tmp/orbit-npm-cache npm pack --dry-run --loglevel=warn

Assessment: The installers now have a deterministic rotation/revocation path with local regression coverage for the signer lifecycle cases requested by ORB-00270.

</details>

## Validation

- Not reported

## Branch Freshness

- Head ref: `orbit/ORB-00270-6a1116cd`
- Behind base: 0
- Ahead of base: 1

### DIFF
--- a/docs/RELEASE.md
+++ b/docs/RELEASE.md
@@ -23,6 +23,19 @@ release workflow signs `orbit-checksums.txt` as `orbit-checksums.txt.sig`;
 both `install.sh` and the npm postinstall authenticate that signature before
 trusting release-hosted SHA-256 values.
 
+The installers carry a small release-signing trust set, not a single forever
+key:
+
+- `orbit-release-2026-05-primary` — current signing path, valid through
+  `2027-12-31`, not revoked.
+- `orbit-release-2026-05-successor` — pre-staged successor signing path,
+  valid through `2028-12-31`, not revoked.
+
+During verification the installers try each known public key, then reject a
+matching key if its `not_after` date has passed or its `revoked_at` field is
+set. A signature that matches none of the trusted keys is rejected as
+untrusted.
+
 ## Steps to cut a release
 
 Each step names the exact file or command. Do them in order.
@@ -129,8 +142,49 @@ Installer environment overrides are trust-boundary changes:
   authenticity, so installers require
   `ORBIT_RELEASE_PUBLIC_KEY_FILE_ACKNOWLEDGE_TRUST_CHANGE=1` and log when the
   override is active.
-- Follow-up task `ORB-00270` tracks key rotation / multi-key support beyond
-  this first pinned release key.
+- `ORBIT_RELEASE_TRUSTED_KEYS_FILE` exists for deterministic installer tests
+  and emergency operations that need a full replacement trust set with key
+  IDs, `not_after`, and `revoked_at` metadata. Each record is
+  `key_id|not_after|revoked_at|public_key_path`; empty `not_after` means no
+  expiry and empty `revoked_at` means active. It also requires
+  `ORBIT_RELEASE_TRUSTED_KEYS_FILE_ACKNOWLEDGE_TRUST_CHANGE=1`.
+
+## Release signing key rotation and revocation
+
+Normal rotation uses an overlap window:
+
+1. Generate the successor keypair offline. Add the public half to the trust
+   set in both `install.sh` and
+   [`plugin/npm/scripts/install-binary.js`](../plugin/npm/scripts/install-binary.js)
+   with a new key ID and `not_after` date. Keep the old active key until at
+   least one npm package containing both keys has been published.
+2. Publish a release and npm package that still signs with the old key, but
+   whose installers trust both old and new keys.
+3. Update `ORBIT_RELEASE_SIGNING_KEY_PEM` and
+   [`plugin/npm/release-signing.pub`](../plugin/npm/release-signing.pub) to the
+   successor key. Cut the next release signed by the successor key.
+4. After the overlap window, remove the old key from the trusted set or set
+   its `revoked_at` date if it should remain visible for audit history.
+
+Emergency revocation is intentionally more disruptive:
+
+1. Mark the compromised key record with `revoked_at: YYYY-MM-DD` in both
+   installers and publish a patch release signed by a non-revoked key.
+2. Update `ORBIT_RELEASE_SIGNING_KEY_PEM` and `release-signing.pub` to the
+   replacement key before cutting that patch release.
+3. Deprecate every already-published npm version whose postinstall still
+   trusts the compromised key, for example:
+
+   ```bash
+   npm deprecate '@orbit-tools/cli@<=X.Y.Z' \
+     'Release signing key revoked; upgrade to a patched @orbit-tools/cli.'
+   ```
+
+   npm packages are immutable after publish. Deprecation warns users and
+   install tooling, but old package tarballs still contain their old trust
+   set. Users pinned to a deprecated package can still run that package's
+   postinstall unless their package manager blocks deprecations, so the
+   release announcement must tell them to upgrade.
 
 Because npm publish is manual, the on-tag smoke run will fail if it fires
 before step 7 completes. That is expected and not actionable on its own;
--- a/docs/RELEASE.md
+++ b/docs/RELEASE.md
@@ -26,15 +26,28 @@ trusting release-hosted SHA-256 values.
 The installers carry a small release-signing trust set, not a single forever
 key:
 
-- `orbit-release-2026-05-primary` — current signing path, valid through
-  `2027-12-31`, not revoked.
-- `orbit-release-2026-05-successor` — pre-staged successor signing path,
-  valid through `2028-12-31`, not revoked.
+- `orbit-release-key-1` — current signing path, valid through `2027-12-31`,
+  not revoked.
+- `orbit-release-key-2` — pre-staged successor signing path, valid through
+  `2028-12-31`, not revoked.
+
+Key IDs are stable generation labels (`key-1`, `key-2`, …). The numeric suffix
+is a generation counter, not a date, so an ID survives the rotation that
+promotes it from successor to primary without becoming confusing.
 
 During verification the installers try each known public key, then reject a
 matching key if its `not_after` date has passed or its `revoked_at` field is
 set. A signature that matches none of the trusted keys is rejected as
 untrusted.
+
+> **Operator custody requirement.** Pre-staging the successor key only buys
+> rotation speed if the successor private key is held in *independent* custody
+> from the primary. If both private halves are stored together (same secrets
+> manager, same machine), a single compromise gets both and the trust set's
+> benefit collapses. The current `orbit-release-key-2` private half is held
+> offline and is only loaded into `ORBIT_RELEASE_SIGNING_KEY_PEM` when the
+> rotation runbook is executed. Future generations must preserve this
+> separation.
 
 ## Steps to cut a release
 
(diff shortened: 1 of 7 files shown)


install.sh
```diff
@@ -69,13 +60,107 @@ pYZDkW2dLqPeFj/WwGhZoYFHv0GOMIWdi6FNriQdkn4RAgMBAAE=
...
+  key_dir="${TMP_DIR}/trusted-release-keys"
+  mkdir -p "$key_dir"
+
+  primary_key_path="${key_dir}/orbit-release-2026-05-primary.pub"
+  successor_key_path="${key_dir}/orbit-release-2026-05-successor.pub"
+  write_release_key_orbit_release_2026_05_primary "$primary_key_path"
+  write_release_key_orbit_release_2026_05_successor "$successor_key_path"
+
+  printf 'orbit-release-2026-05-primary|2027-12-31||%s\n' "$primary_key_path"
+  printf 'orbit-release-2026-05-successor|2028-12-31||%s\n' "$successor_key_path"
```
danieljhkim: Ops question this PR should answer: **where is the private half of `orbit-release-2026-05-successor`?**

If it lives in the same secrets manager that holds the primary signing key, a single compromise gets both and the trust set's benefit collapses. The whole point of a pre-staged successor is that its public half ships now while the private half is held in cold storage (HSM, sealed envelope, offline machine) until a rotation event. The RELEASE.md text says "generate the successor keypair offline" which is right for *future* rotations, but doesn't explain the custody of *this* successor key, which has presumably already been generated.

Worth either calling out the custody answer in the PR description, or filing a follow-up to document it in the runbook before this trust set goes into a real release.

danieljhkim (commented):
Good follow-up on the rotation comment from #410. The two-key trust set with `not_after` + `revoked_at`, the matching guardrail expansion to require ≥2 keys + canonical key + known IDs, and the expired/revoked/untrusted shell+npm test coverage all match what I'd want to see for this AC. The RELEASE.md write-up of normal-rotation vs emergency-revocation procedures is clear and the npm-immutability caveat is honestly stated.

Things to address or at least answer before merge:

- **Where is the successor private key right now?** The whole point of a pre-staged successor is that the public half ships now and the private half is held in cold storage until rotation. If both private halves live in the same secrets manager / on the same machine, a single compromise gets both and the rotation gives you nothing. Not a code question; an ops question the PR description should answer.
- **Naming.** `orbit-release-2026-05-successor` is going to read as misleading once it's been the primary for a year. Suggest `orbit-release-key-1` / `key-2`, or use the SPKI fingerprint as the ID.
- **Two redundant override mechanisms.** `ORBIT_RELEASE_PUBLIC_KEY_FILE` is now strictly a subset of `ORBIT_RELEASE_TRUSTED_KEYS_FILE`. Worth at least flagging the old one as deprecated-in-favor-of in docs, so we don't carry both forever.
- A few smaller items inline.

install.sh
```diff
@@ -69,13 +60,107 @@ pYZDkW2dLqPeFj/WwGhZoYFHv0GOMIWdi6FNriQdkn4RAgMBAAE=
...
+    { exit 1 }
+  ' || fail "invalid release signing key date: $value"
+}
+
+write_builtin_trusted_key_records() {
+  key_dir="${TMP_DIR}/trusted-release-keys"
+  mkdir -p "$key_dir"
+
+  primary_key_path="${key_dir}/orbit-release-2026-05-primary.pub"
+  successor_key_path="${key_dir}/orbit-release-2026-05-successor.pub"
```
danieljhkim: Nit on naming: `orbit-release-2026-05-successor` reads naturally today but will be confusing once it's actually been the primary key for a year. When the next rotation introduces a *third* key in 2027, the operator will need to remember that "the 2026-05-successor" is the *current* primary, while the new key is *its* successor.

Consider stable nonce-style IDs instead: `orbit-release-key-1` / `key-2`, or the SPKI fingerprint as the ID. The `not_after` and `revoked_at` fields already encode the lifecycle; the ID doesn't need to.

plugin/npm/scripts/install-binary.js
```diff
@@ -20,10 +20,50 @@ const BIN_DIR = path.join(PKG_ROOT, 'binaries');
...
+const TRUSTED_KEYS_OVERRIDE = process.env.ORBIT_RELEASE_TRUSTED_KEYS_FILE;
+const TRUSTED_KEYS_OVERRIDE_ACK_ENV = 'ORBIT_RELEASE_TRUSTED_KEYS_FILE_ACKNOWLEDGE_TRUST_CHANGE';
 const TRUSTED_PUBLIC_KEY_PATH = PUBLIC_KEY_OVERRIDE
   ? path.resolve(PUBLIC_KEY_OVERRIDE)
   : path.join(PKG_ROOT, 'release-signing.pub');
+const TRUSTED_KEYS_OVERRIDE_PATH = TRUSTED_KEYS_OVERRIDE ? path.resolve(TRUSTED_KEYS_OVERRIDE) : null;
+const TRUSTED_RELEASE_KEYS = Object.freeze([
+  Object.freeze({
+    id: 'orbit-release-2026-05-primary',
+    notAfter: '2027-12-31',
```
danieljhkim, line 32: The shell side validates `not_after` format via awk regex (`release_date_number`). The JS side compares with `releaseDateString(verificationDate) > key.notAfter` and trusts whatever string is in `notAfter`.

If a malformed override sets `notAfter: "next month"`, the comparison `"2026-05-23" > "next month"` is false → the key is silently treated as never-expiring. Same for `notAfter: ""` (your `|| null` handles that, so OK there), but `notAfter: "foo"` slips through.

Quick fix: add `if (!/^\d{4}-\d{2}-\d{2}$/.test(key.notAfter)) throw new Error(...)` in `normalizeTrustedReleaseKeys` for any non-null `notAfter`. Mirror what the shell does.

plugin/npm/scripts/install-binary.js
```diff
@@ -87,6 +127,9 @@ function parseChecksums(text) {
 }
 
 function acknowledgeTrustedPublicKeyOverride() {
+  if (PUBLIC_KEY_OVERRIDE && TRUSTED_KEYS_OVERRIDE) {
+    throw new Error('ORBIT_RELEASE_PUBLIC_KEY_FILE and ORBIT_RELEASE_TRUSTED_KEYS_FILE cannot both be set');
```
danieljhkim, line 131: `ORBIT_RELEASE_PUBLIC_KEY_FILE` and `ORBIT_RELEASE_TRUSTED_KEYS_FILE` are now overlapping override mechanisms — a trust-keys manifest with a single row is equivalent to the public-key file (sans expiry/revocation metadata).

The mutual-exclusion check is correct, but two override surfaces means twice the docs, twice the failure modes, and twice the env vars to remember to scrub in tests. Consider flagging `ORBIT_RELEASE_PUBLIC_KEY_FILE` as deprecated in favor of the new one (with `log()` line on use) and dropping it in the v0.7 release. The new mechanism is strictly more expressive; the old one is just a less-capable subset now.

scripts/check-installer-pubkey.sh
```diff
@@ -3,28 +3,74 @@ set -euo pipefail
...
+if [ "$npm_key_count" -lt 2 ]; then
+  echo "check-installer-pubkey: expected at least two public key blocks in install-binary.js, found $npm_key_count" >&2
   exit 1
 fi
 
+assert_contains_canonical_key "install.sh" "$install_keys"
+assert_contains_canonical_key "install-binary.js" "$npm_keys"
+
+grep -q "orbit-release-2026-05-primary" "$repo_root/install.sh"
+grep -q "orbit-release-2026-05-successor" "$repo_root/install.sh"
```
danieljhkim: These `grep -q "orbit-release-2026-05-*"` assertions will break on every legitimate rotation until someone remembers to update them. That's the opposite of what you want — a guardrail that triggers on the operation it's supposed to safeguard.

Durable alternative: read the key IDs out of `install-binary.js`'s `TRUSTED_RELEASE_KEYS` array (it's structured data), then assert that every ID appears in `install.sh` and vice versa. Or generate `install.sh`'s key list from the JS object at build time. As-is, this script is a tripwire that goes off on rotations *and* requires hand-editing each time.

docs/RELEASE.md
```diff
@@ -129,8 +142,49 @@ Installer environment overrides are trust-boundary changes:
...
+   replacement key before cutting that patch release.
+3. Deprecate every already-published npm version whose postinstall still
+   trusts the compromised key, for example:
+
+   ```bash
+   npm deprecate '@orbit-tools/cli@<=X.Y.Z' \
+     'Release signing key revoked; upgrade to a patched @orbit-tools/cli.'
+   ```
+
+   npm packages are immutable after publish. Deprecation warns users and
```
danieljhkim: This paragraph is the most important sentence in the doc and it's buried inside a code block. Worth lifting it into a callout / warning block:

> **Emergency revocation only protects users who upgrade.** Already-published npm packages contain their old trust set permanently. A revoked key in the *current* trust set blocks new releases signed by the compromised key — it does **not** retroactively block users from running already-published `@orbit-tools/cli@<old>` packages. The release announcement and `npm deprecate` are the only revocation mechanisms for those installs.

Readers who skim will otherwise come away thinking "revocation is supported" without internalizing the immutability caveat. This is the kind of thing a security review will flag later.

PR #501, opened 2026-06-13
refactor: remove the Shell activity variant and run_shell dispatch [ORB-00374]

danieljhkim:
## Summary

Removes the v2 `Shell` activity surface end to end — the fail-closed resolution of security bug **ORB-00363**.

`run_shell` spawned `Command::new(program)` with **no OS sandbox, no cwd confinement, and no policy consultation** (unlike the agent-CLI path), and its program allowlist was self-asserted within the same workspace-supplied spec — both `program` and `allowed_programs` came from the same YAML, so the gate was a tautology. A workspace activity/job declaring `type: shell` now **fails to deserialize at load instead of executing**.

## What changed

- **Type surface:** dropped `ActivityV2Spec::Shell` + `ShellSpec`, the `ShellProgramNotAllowed` / `ShellSpawnFailed` / `ShellExitedUnexpected` `DispatchError` variants, and every cross-crate match arm / re-export that named them.
- **Dispatch:** removed `run_shell`, its dispatch arm, the `"shell"` label, and the now-dead `std::process::Command` import.
- **Demos (per chosen scope):** deleted 7 pure-shell job fixtures, the `shell_reference` asset, and the shell smoke legs in both example binaries.
- **Tests migrated to `deterministic`:** JobKind, round-trip, and workspace-shadow tests. The shadow test now uses an *unregistered* action, so an incorrectly-executed shadow **fails the run** rather than writing a marker — same security assertion.
- **Docs:** `write-activity.md` and `activity-job-yaml.md` updated to drop the shell type.

## Coverage note

Two shell-specific tests asserting the `Ok(success=false)` audit-message path were removed — deterministic actions cannot produce that outcome, so that path's only coverage was shell-based. Flagged in the task execution summary.

## Verification

- workspace + examples build clean
- `orbit-common` / `orbit-engine` / `orbit-cli` / `orbit-core` targeted tests green
- `make ci-fast` passed (all guardrail scripts)

Resolves ORB-00363.


danieljhkim:
Follow-up commit `bb7692d0` addresses the two review findings:

**1. Stale design docs (blocker).** The original commit updated the website how-to/reference docs but not the feature design docs under `docs/design/activity-job/`, which still listed `shell` as a live activity type.
- `1_overview.md`, `2_design.md`: dropped `shell` from the activity-type surface, the dispatch delegation table, and the non-retryable error list; added a one-line pointer to the removal where a reader would otherwise expect the variant.
- `4_decisions.md`: added **ADR-0194** (proposed) recording the fail-closed removal as the security decision behind ORB-00363 — context, decision, and the two costs (lost `Ok(success=false)` coverage; no compat shim, old YAML fails at load). Allocated via `orbit.adr.add`; artifact committed under `.orbit/adrs/proposed/ADR-0194/`.
- Bumped `last_updated` and added ORB-00363/ORB-00374 task references in all three.

**2. Dead-code cleanup.** Five doc comments the removal commit missed still named the `shell` variant (the `activity_v2.rs` discriminator example, `job_v2.rs` retry note, orbit-core `activity_v2.rs` backend note, `dispatcher.rs` host note, `step.rs` retry note) — all fixed, with the retained `Ok(success=false)` machinery rationale re-anchored to ADR-0194.

Note: the `DispatchOutcome.success` field is **kept** — it's load-bearing across parallel/fan-out/loop/recovery aggregation, not shell-only scaffolding, so removing it would be a risky cross-cutting refactor outside this PR's scope. ADR-0194's second consequence documents that it's now retained as the general block-level outcome contract.

Comment- and doc-only; no behavior change. `make ci-fast` passes. ADR-0194 should flip `proposed → accepted` (`orbit.adr.update`) when this merges.

PR #558, opened 2026-07-11
Document Orbit operational log diagnostics

danieljhkim:
## Summary

- Add a workflow-skill reference for investigating Orbit service, JSONL, and job-run logs.
- Link the new reference from the bundled `orbit-workflow` skill.

## Why

On Linux, `orbit sweep` writes to the journal but still attempts to prune the macOS-only sweep-log path. If that directory is absent, the shared rotation helper emits a non-fatal ENOENT warning every minute. The reference explains how to distinguish this warning from a real retention or job-run failure.

## Impact

Documentation only; no runtime behavior or host configuration changes.

## Validation

- `make ci-fast`
- `git diff --check`

### DIFF
--- a/crates/orbit-core/assets/skills/orbit-workflow/references/operational-logs.md
+++ b/crates/orbit-core/assets/skills/orbit-workflow/references/operational-logs.md
@@ -1,0 +1,126 @@
+# Checking Orbit Operational Logs
+
+Use this reference for host-level incidents and warnings: an `orbit-sweep`
+service failure, a global JSONL tracing warning, a missing log file, or a
+question about which evidence to inspect. For a specific failed job run, start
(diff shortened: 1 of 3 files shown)


crates/orbit-core/assets/skills/orbit-workflow/references/operational-logs.md
```diff
@@ -0,0 +1,126 @@
...
+orbit sweep --dry-run
+```
+
+## Global JSONL Tracing
+
+The JSONL tracing sink is distinct from the sweep-service log. Its active file
+and rename-based archives are disposable; job/run stores are not.
+
+```bash
+find "$orbit_root/state/logs" -maxdepth 1 -type f -printf '%f %s bytes\n' | sort
```
danieljhkim, line 73: why not use `orbit log tail/audit`

PR #581, opened 2026-07-12
feat: Record managed-skill ownership and retirement metadata [ORB-10181]

danieljhkim:
## Summary

Adds persistent managed-skill ownership + retirement metadata and tightens unlink/teardown to it (task **ORB-10181**).

- New `crates/orbit-core/src/command/skill_ownership.rs`: `SkillOwnershipManifest` (schema v1) persisted at `<skills_root>/.ownership.json` — `managed` records (skill id, sha256 generated-content hash, catalog version, owned root, managed link destinations) + `tombstones` (retired ids with retained hashes/version/roots/links).
- `reconcile_managed_skills()`: upserts a record per current generated skill, retires vanished ones into tombstones, parks a superseded hash on template-change bumps so pre-upgrade installs stay provable; idempotent (skips write when unchanged).
- `classify_install()` / `classify_symlink()`: `OrbitOwned` only via exact known hash (managed+tombstone) OR a symlink resolving inside an owned root; modified generated content → `Ambiguous`, unknown id → `Unmanaged`.
- `remove_owned_skill_links()`: removes only proven-owned symlinks, leaving user symlinks and regular files/dirs intact.
- Wiring in `init.rs`, `skill.rs`, `teardown.rs`; removed the now-dead `support::remove_symlinks_in` helper. CHANGELOG + gc design-doc `related_artifacts`.

## Tests

- `crates/orbit-core/src/command/tests/skill_ownership.rs` — 12 tests covering current generated, retired generated (hash-owned via tombstone), template-change hash retention, modified copy (ambiguous), same-named user dir (unmanaged), symlink-into-owned-root, broken link into owned root, symlink outside roots (ambiguous), owned-only link removal, link-destination recording, corrupt manifest, save/load round-trip.
- Validation this rescue: `cargo test -p orbit-core --lib` → **554 passed, 0 failed**; focused `skill_ownership` → 12 passed; `git diff --check` clean.
- Note: `make ci-fast`'s `scripts/test-installer-security.sh` needs the `node` binary (absent in this environment); the CI merge gate runs it unrestricted.

## Rescue provenance

Pipeline run `jrun-20260712-2131-4` (crew opus) implemented + tested + committed successfully, but the deterministic `pr_open` step failed because the task branch was behind `origin/agent-main`. The branch was never pushed, so the work lived only in the managed pipeline worktree. This PR carries that same work (head `c860fedf`), rebased onto current `agent-main` (`dff4fd76`); no integration conflicts. The Orbit task is intentionally left un-transitioned.

[ORB-10181]


crates/orbit-core/src/command/skill_ownership.rs
```diff
@@ -0,0 +1,437 @@
...
+            tombstones: BTreeMap::new(),
+        }
+    }
+}
+
+/// Ownership metadata for one currently-managed skill.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub struct ManagedSkillRecord {
+    pub skill_id: String,
+    /// sha256 (hex) of the generated `SKILL.md` as rendered for `owned_root`.
```
danieljhkim: **[Spec compliance] — Ownership hash ignores generated resources and other tree changes.**

The manifest hashes only rendered `SKILL.md`, but `seed_default_skill_resources` writes additional generated files. If a user modifies a generated resource or adds a file while leaving `SKILL.md` unchanged, `classify_install` still returns `OrbitOwned`; stale-skill GC could then remove modified content, violating the requirement that modified generated skills are never automatically claimed or removed.

Use a deterministic no-follow fingerprint of the complete generated skill tree (paths, file types, and contents, with explicit exclusions for ownership metadata), retain those historical fingerprints in tombstones, and add tests for modified resources plus added files.

danieljhkim: Addressed in 9deecd0d.

The `SKILL.md`-only hash is replaced with a deterministic, no-follow **whole-tree fingerprint** of every Orbit-generated file for a skill.

**Encoding (documented on `fingerprint` in `skill_ownership.rs`).** `sha256` over a domain tag then, for every entry sorted by `(encoded-path, kind-tag)`: the LE-`u64` path length, the path bytes, a one-byte kind tag (`d`/`f`/`l`/`o`), and a 32-byte content digest (file contents' sha256, all-zero for non-files). Paths are skill-root-relative and joined with `/` on every platform; the length prefix makes the concatenation injective. Directories are entries (so an added empty dir is caught); symlinks and exotic node types are recorded **by type only and never followed**.

**Symmetry.** Seeding (`record_default_skill_ownership` → `GeneratedSkill::from_files`) computes the *expected* fingerprint over the compiled-in generated set (`SKILL.md` + matching `references/*`); `classify_install` computes the *actual* fingerprint by walking the on-disk tree with `symlink_metadata` (no-follow), then matches it against known **managed + tombstone** fingerprints. Retirement and template/resource bumps park the prior fingerprint on the tombstone, so historical fingerprints stay queryable.

**Fail-closed.** Modified resource, added file, removed generated file, embedded symlink, and unexpected entry types now all change the fingerprint → `Ambiguous`; symlinks are never traversed, so a link cannot redirect the walk outside the skill root. The `.ownership.json` manifest is excluded structurally (it sits at the skills root, outside every skill tree), so the fingerprint is not self-referential.

**Tests** (`tests/skill_ownership.rs`, 18 total): added `modified_resource_is_ambiguous`, `added_file_is_ambiguous`, `removed_generated_file_is_ambiguous`, `embedded_file_symlink_is_not_followed_and_is_ambiguous` (target holds the exact generated bytes — proves no-follow), `embedded_directory_symlink_is_not_followed_and_is_ambiguous`, `current_generated_tree_with_resources_is_orbit_owned`, and `resource_change_keeps_previous_fingerprint_as_tombstone`; all prior ownership/link/tombstone tests preserved.

**Validation:** `cargo test -p orbit-core --lib` → 560 passed, 0 failed; focused `skill_ownership` → 18 passed; `cargo clippy -p orbit-core --lib` clean; `cargo fmt --check` clean; `git diff --check` clean; changelog style/freshness guardrails pass.

PR #577, opened 2026-07-12
Pipeline worktree GC: reclaim orphaned jrun worktrees

danieljhkim:
## Task

[ORB-10173](https://orbit-cli.com/tasks/ORB-10173) — Pipeline worktree GC: reclaim orphaned jrun worktrees

## Problem

A disk-cleanup sweep on dk-server-1 (2026-07-12) found **116G** of stale Orbit pipeline job-run worktrees under `codebases/orbit/.orbit/state/worktrees/orbit-jrun-*/` — 13 directories, each 519M–20G, almost entirely build artifacts. No run was active against any of them; they are orphaned leftovers of completed/failed job runs.

Root fs was at 76% (248G/342G). The sweep reclaimed 113G from *unmanaged* target dirs (primary checkout + dev worktrees) but deliberately did not touch anything under `.orbit/state/` — that is Orbit-managed state, and reclaiming it is Orbit's job, not a shell script's.

Orbit creates a job worktree per pipeline run and never reclaims it, so disk grows every run and never shrinks. This is a slow leak that will re-fill the disk within weeks.

Related: friction **F2026-07-019** (filed by the cleanup run), same orphaned-run family as **F2026-07-016 / ORB-10153**.

## Scope — two parts

**1. Worktree GC on run completion.** When a job run reaches a terminal state (done / failed / cancelled), Orbit should remove its worktree. Needs a retention policy — a failed run's worktree is useful for debugging, so keep failures for N days (suggest 7) or the last K failures, and reap successes immediately (or after a short grace window). Make the policy configurable, with sane defaults.

**2. Reclaim the existing orphans + reconcile.** GC-on-completion doesn't help worktrees whose run record is already terminal-or-lost. Add a sweep/`gc` entry point that reconciles `.orbit/state/worktrees/*` against the run table and reclaims any worktree with no live run (cancelling orphaned run records — `orbit run cancel <jrun>` — where the record is stuck non-terminal). It must be safe to run while other runs are in flight: never touch a worktree with a live run, and hold whatever lock the pipeline uses.

GC must be **language-agnostic** — it reclaims the worktree directory, whatever the workspace's build system put there (Rust `target/`, Java `build/`, Python `.venv/`, node_modules, …). Nothing here should special-case Cargo.

## Out of scope

[...]

PR #582, opened 2026-07-12
Add managed pipeline worktree GC under orbit gc

danieljhkim:
## Task

[ORB-10182](https://orbit-cli.com/tasks/ORB-10182) — Add managed pipeline worktree GC under orbit gc

## Problem

Pipeline worktrees and their language-specific build artifacts can accumulate without bound. PR #577 / ORB-10173 attempted cleanup under `orbit run gc`, but its command ownership and concurrency/data-safety model were insufficient.

## Why It Matters

Worktrees are the largest known disk leak, but they may also contain live work, resumable failures, dirty changes, or unmerged branches. Collection must be urgent without being lossy.

## Scope

Implement `orbit gc worktrees` on the shared framework. Reconcile only Orbit-managed worktrees with run ownership, reuse the classifier for terminal-run cleanup, and apply status-specific retention.

## Constraints / Notes

- Never delete a live/current/inconclusive worktree.
- Revalidate ownership atomically under the GC/run-claim lock immediately before mutation.
- Do not blindly delete unknown directories, dirty source-bearing trees, or unmerged branches.
- Removal is language-neutral and must not special-case Cargo.

### Acceptance Criteria

- `orbit gc worktrees` reports and, only with `--apply`, reclaims eligible Orbit-managed worktrees using the shared GC report.
- Live-run, current-cwd, liveness-inconclusive, dirty source-bearing, unknown-owner, and resumable interrupted worktrees are retained with explicit reasons.
- Success/cancelled and failed/interrupted worktrees have separately configurable retention based on persisted run timestamps.
- Ownership is revalidated atomically against run claim/start before deletion; a race test proves a pending run becoming live cannot be cancelled or removed.
- Use `git worktree remove/prune`; never automatically delete an unmerged or unpushed task branch.
- On-terminal cleanup and manual GC share exactly the same classifier and retention policy.
- Tests cover live, dead, missing-record, current-cwd, dirty, retained failure, unknown directory, and idempotent/concurrent execution.
- A live-box validation records before/after worktree bytes and filesystem usage, and resolves or explicitly re-scopes F2026-07-019 plus the overlapping ORB-10153 family.
- `make ci-fast` and focused core/CLI tests pass.

## Execution Summary

<details>

Outcome: success
Changes:
- Added the `WorktreeGcCollector` to the shared plan/apply/report framework with persisted run clocks, separate success/failure retention, current-cwd/live-or-inconclusive/resumable/dirty/unknown-owner/unmerged/unpushed protections, Git-only removal/prune, immediate revalidation, byte accounting, and idempotent manifests.
- Wired `orbit gc worktrees`, qualified CLI retention overrides, selected-workspace runtime resolution, typed `[gc.worktrees]` config, config reporting, and design/changelog documentation.
- Routed terminal-run cleanup through the same collector and changed the engine cleanup action into a non-destructive defer, eliminating force removal, raw deletion, and automatic branch deletion.
- Added classifier, real-Git plan/apply, pending-to-live race, dirty/missing/unknown/current/retained/resumable/branch-safety, idempotency, config, and CLI tests. Added corrective learning L-0080; linking it as the superseder of remote-only L-0078 was blocked by the lifecycle tool and tracked as F2026-07-026.
- Live-box apply scanned 21 worktrees and reclaimed 14 / 138,582,642,381 estimated bytes with a clean outcome; `.orbit/state/worktrees` fell from 163,801,865,718 to 43,333,830,541 bytes and root filesystem usage from 47% to 16%. Seven protected trees remained with explicit reasons. This explicitly re-scopes F2026-07-019 and the ORB-10153 overlap to safety-first ownership.
Assessment: Safety invariants are fail-closed, manual and terminal cleanup share one collector, branches are never deleted, and focused behavior is covered against real Git worktrees.
Validation:
- PASS: cargo check -p orbit-core -p orbit-cli
- PASS: 6 focused orbit-core worktree GC tests
- PASS: worktree GC config test
- PASS: 5 focused orbit-cli GC tests
- PASS: cargo fmt --all -- --check; git diff --check
- PASS: all remaining ci-fast guardrails (dependency direction, CLI imports, stability, learning layout, redaction, changelog freshness/style, error translation)
- PARTIAL: make ci-fast stopped at scripts/test-installer-security.sh because `node` is not installed on PATH; preceding fmt/pubkey checks passed, and all subsequent guardrails were run individually and passed.

</details>

## Validation

- Not reported

## Branch Freshness

- Head ref: `orbit/ORB-10182-6a540bbf`
- Behind base: 0
- Ahead of base: 1

### DIFF
--- a/crates/orbit-core/src/command/gc/worktrees.rs
+++ b/crates/orbit-core/src/command/gc/worktrees.rs
@@ -1,0 +1,976 @@
+//! Managed pipeline-worktree collector for the shared `orbit gc` framework.
+
+use std::fs;
+use std::path::{Path, PathBuf};
+use std::process::{Command, Output, Stdio};
+
+use chrono::{DateTime, Duration, Utc};
+use orbit_common::types::{JobRun, JobRunState, OrbitError};
+use serde::{Deserialize, Serialize};
+
+use crate::OrbitRuntime;
+
+use super::{
+    GcCandidate, GcCollector, GcContext, GcItemError, GcMutation, GcPlan, GcRequest,
+    GcRevalidation, GcScope, GcSkip, GcTarget, SystemGcClock, execute_gc,
+};
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct WorktreeGcPolicy {
+    pub success_retention_days: u64,
+    pub failure_retention_days: u64,
+}
+
+impl WorktreeGcPolicy {
+    pub fn from_runtime(runtime: &OrbitRuntime) -> Self {
+        Self {
+            success_retention_days: runtime.worktree_gc_success_retention_days(),
+            failure_retention_days: runtime.worktree_gc_failure_retention_days(),
+        }
+    }
+}
+
+pub struct WorktreeGcCollector<'a> {
+    runtime: &'a OrbitRuntime,
+    policy: WorktreeGcPolicy,
+    only_run_id: Option<String>,
+}
+
+impl<'a> WorktreeGcCollector<'a> {
+    pub fn new(runtime: &'a OrbitRuntime, policy: WorktreeGcPolicy) -> Self {
+        Self {
+            runtime,
+            policy,
+            only_run_id: None,
+        }
+    }
+
+    pub(crate) fn for_run(
+        runtime: &'a OrbitRuntime,
+        policy: WorktreeGcPolicy,
+        run_id: &str,
+    ) -> Self {
+        Self {
+            runtime,
+            policy,
+            only_run_id: Some(run_id.to_string()),
+        }
+    }
+
+    fn worktree_root(context: &GcContext<'_>) -> PathBuf {
+        context.scope.root().join("state").join("worktrees")
+    }
+
+    fn classify_path(
+        &self,
+        path: &Path,
+        context: &GcContext<'_>,
+    ) -> Result<Classified, OrbitError> {
+        let id = path
+            .file_name()
+            .and_then(|name| name.to_str())
+            .unwrap_or("<non-utf8>")
+            .to_string();
+        let Some(run_id) = extract_run_id(path) else {
+            return Ok(Classified::skip(
+                id,
+                "unknown_directory",
+                "directory is not named for an Orbit job run",
+            ));
+        };
+        if self
+            .only_run_id
+            .as_deref()
+            .is_some_and(|only| only != run_id)
+        {
+            return Ok(Classified::Ignored);
+        }
+        if current_worktree_contains(path) {
+            return Ok(Classified::skip(
+                id,
+                "current_worktree",
+                "worktree contains the current process directory",
+            ));
+        }
+        let Some(run) = self.runtime.stores().jobs().get_run(&run_id)? else {
+            return Ok(Classified::skip(
+                id,
+                "unknown_owner",
+                "no persisted run record proves ownership of this worktree",
+            ));
+        };
+        let Some(git) = inspect_git_worktree(
+            &self.runtime.paths().repo_root,
+            path,
+            self.runtime.workflow_base_branch(),
+        )?
+        else {
+            return Ok(Classified::skip(
+                id,
+                "unknown_owner",
+                "directory is not registered by Git as a managed worktree",
+            ));
+        };
+        if git.dirty {
+            return Ok(Classified::skip(
+                id,
+                "dirty_source",
+                "worktree has tracked or untracked source-bearing changes",
+            ));
+        }
+        if let Some(branch) = git.branch.as_deref() {
+            if !git.merged {
+                return Ok(Classified::skip(
+                    id,
+                    "unmerged_branch",
+                    format!("branch `{branch}` is not merged into the configured base"),
+                ));
+            }
+            if is_task_branch(branch) && !git.pushed {
+                return Ok(Classified::skip(
+                    id,
+                    "unpushed_branch",
+                    format!("task branch `{branch}` is not present on a remote"),
+                ));
+            }
+        }
+        match classify_run(
+            &run,
+            self.policy,
+            context.clock.now(),
+            self.interrupted_is_resumable(&run.run_id)?,
+        ) {
+            RunClassification::Skip { code, reason } => Ok(Classified::skip(id, code, reason)),
+            RunClassification::Eligible { retention_evidence } => {
+                let bytes = directory_bytes_no_follow(path).ok();
+                let expected = ExpectedState {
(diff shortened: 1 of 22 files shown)


crates/orbit-core/src/command/gc/worktrees.rs
```diff
@@ -0,0 +1,976 @@
...
+                "candidate path is absent",
+            ));
+        };
+        if current_worktree_contains(path) {
+            return Ok(skip_revalidation(
+                "current_worktree",
+                "worktree became the current process directory",
+            ));
+        }
+        let Some(run) = self.runtime.stores().jobs().get_run(&expected.run_id)? else {
```
danieljhkim, line 216: **[Spec compliance] — Atomic run-owner revalidation is not held through deletion.**

`revalidate_candidate` reads the run row and Git state, returns `Ready`, and then `apply` performs `git worktree remove` later without a run-claim/store lock spanning the two operations. It also never checks the persisted PID/start identity for a terminal row, so a still-live or inconclusive owner can be treated as eligible. The current pending-to-running test changes state before a fresh apply plan; it does not race a frozen candidate between final validation and removal.

Hold a run-claim/store transaction or equivalent ownership guard through removal, fail closed on PID/start-identity uncertainty, and add a deterministic frozen-plan race test.

danieljhkim: Fixed in 5c744eeb.

**Atomic revalidation through removal.** `apply` now re-runs the full owner state/identity/liveness + Git revalidation (`revalidate_candidate`) as the immediately preceding operation to `git worktree remove`, with no intervening store write or fs sync, under the host-global GC lock that `execute_gc` already holds across apply. If the owner is claimed/transitioned between planning and mutation, apply fails closed rather than removing. The frozen candidate also carries the owner PID + start-identity, so a claim/takeover that rewrites either is rejected as `owner_changed`.

**PID + start-identity, fail-closed, including terminal rows.** Both plan classification and revalidation now call `gc_owner_permits_reclaim`, which reuses run reconciliation's PID + process-start-identity probe (`classify_run_owner`). A row that reads terminal while its worker is still alive (e.g. 0-day success retention racing finalize) is retained. Only a conclusively dead owner (`Missing`/`Mismatch`) — or this same collecting process (terminal cleanup; cwd guard covers the in-use case) — permits removal; verified-live, legacy-live, and inconclusive (`ProbeUnavailable`) probes all fail closed.

**Deterministic race test.** Replaced the insufficient replan-after-transition coverage with two tests:
- `terminal_run_with_live_owner_is_retained` — a **terminal** (Success) row whose recorded PID + start-identity is a live process is skipped `live_or_inconclusive` (0 eligible), proving terminal-row liveness fail-closed.
- `frozen_candidate_is_not_removed_when_owner_claimed_before_mutation` — plans an eligible candidate, **freezes** it, then claims the owner live between planning and mutation; `revalidate` skips and the `apply` atomic guard fails closed, so the worktree survives and the run is left exactly as the claimer set it (not cancelled, not removed).

**Evidence:** `cargo test -p orbit-core command::gc::worktrees` 8/8 pass; `command::job::run` 26/26 pass (owner-change safety unaffected); CLI `gc_*` 5/5 pass; `cargo clippy -p orbit-core -p orbit-cli --all-targets` clean; `cargo fmt --all --check` and `git diff --check` clean; changelog style/freshness guards pass.

**Residual note:** terminal state is monotonic (the state machine forbids terminal→running), so the realistic live-owner path on a terminal-classified candidate is the liveness dimension, which is probed directly microseconds before removal. Fully eliminating the sub-instruction store-read TOCTOU would require a shared kernel lock across the run-claim path; `git worktree remove` (no `--force`) refusing an in-use/dirty tree is an additional backstop. Non-Unix keeps the pre-liveness behavior (no probe available) and relies on the terminal-state/cwd/dirty/branch gates.

Left unresolved for your verification.

danieljhkim: The added PID/start-identity checks and frozen-candidate test are good, but the blocker is not fully closed. `apply` still performs a read-only revalidation and then starts an external `git worktree remove` with no lock/transaction shared with the run-claim path; "adjacent instructions" under the GC-only lock are not atomic against a concurrent claimer. The worker summary itself records this residual TOCTOU. Please introduce a per-run claim/reclaim lock or store transaction/CAS guard that the claim path and GC both honor, hold it from final validation through removal, and race the actual competing claim against that guard.

danieljhkim: Addressed in a7fb5f2c — the revalidate-then-remove window is now closed by a shared per-run claim guard, not just adjacency under the GC-only lock.

**Shared per-run guard.** New `runtime::run_claim_guard`: one advisory file lock (`fs2`) keyed by run id under `state/run-guards/<id>.lock`, derived identically by both paths from the workspace `state` dir so they rendezvous on the same inode (empty / separator / `..` run ids are rejected). It is a filesystem lock, never the SQLite write lock — no unrelated DB lock is held across the git op.

**Both paths honor it.** The run claim/start path holds the guard across its ownership transition — `mark_job_run_running`, `claim_pending_job_run_owner`, `take_over_running_job_run` in `store_delegates.rs` all `acquire(...)` before the store mutation. The worktree collector (`worktrees.rs` `apply`) recovers the owning run id from the frozen plan and holds the *same* guard continuously across `[final revalidation .. git worktree remove]`, with no intervening store write or fs sync. Lock ordering is fixed: **host GC lock (ADR-0220) → per-run guard → filesystem mutation.**

**Mutual exclusion, both interleavings raced against the real guard:**
- `apply_blocks_on_a_claimant_held_run_guard_and_fails_closed` — a claimant grabs the guard first and records a live owner (takeover); `apply` provably blocks on the guard (GC thread only starts after the claimant holds it), and once it wins the guard after release, revalidation observes the live owner and fails closed — the worktree survives.
- `claimant_blocks_on_a_gc_held_run_guard_and_re_evaluates_after_removal` — the mirror: GC holds the guard across the real `git worktree remove` + `prune`; the concurrent claimant blocks until release and only then re-evaluates, never entering a removed tree.

Together these prove the collector genuinely acquires the guard and holds it from validation through removal, and that a concurrent claim contends on that same guard rather than racing it.

**Evidence (recovery run):** `cargo test -p orbit-core --lib command::gc::worktrees` 10/10; `command::job::run` 26/26 (owner-change safety unaffected); `cargo test -p orbit-cli gc` 5/5; `cargo clippy -p orbit-core -p orbit-cli --all-targets` clean; `cargo fmt --all --check` and `git diff --check` clean.

Left unresolved for your verification.

PR #584, opened 2026-07-12
Add unified audit retention and blob mark-and-sweep

danieljhkim:
## Task

[ORB-10186](https://orbit-cli.com/tasks/ORB-10186) — Add unified audit retention and blob mark-and-sweep

## Problem

Existing `orbit audit prune` deletes only legacy audit-event rows. Current audit evidence also includes workspace v2 events, v2 loop JSONL envelopes, and content-addressed blobs that may be shared across retained traces.

## Why It Matters

Audit storage grows continuously, but age-deleting files or blobs can make retained job runs non-diagnosable and corrupt referential integrity.

## Scope

Implement `orbit gc audit` across legacy and v2 audit surfaces, including reference-based blob mark-and-sweep and a compatibility path from `orbit audit prune`.

## Constraints / Notes

- Blobs are deleted by reachability across all retained envelopes, never age alone.
- Interrupted apply must be restart-safe.
- GC must leave an audit record or signed deletion manifest without recursive log growth.

### Acceptance Criteria

- `orbit gc audit` explicitly covers legacy `audit_events`, workspace v2 events, v2 loop JSONL, and content-addressed audit blobs.
- Blob candidates are computed by mark-and-sweep across every retained reference; shared/referenced blobs are never deleted.
- Configured holds/exports and evidence required by retained job runs prevent pruning.
- Existing `orbit audit prune` delegates to or emits a clear deprecation path toward the same collector semantics.
- Apply ordering is transactional/restart-safe and cannot leave retained envelopes pointing at missing blobs.
- The GC operation records its own deletion manifest/audit event without recursive collection.
- Tests cover shared blobs, missing blobs, malformed JSONL, concurrent writers, holds, retained-run evidence, cross-workspace isolation, interrupted restart, and idempotency.
- `make ci-fast` and focused audit/store/CLI tests pass.

## Execution Summary

<details>

Outcome: success
Changes:
- Added a workspace-scoped `AuditGcCollector` covering legacy audit rows, workspace v2 rows, loop JSONL envelopes, holds/exports, retained job-run evidence, and content-addressed blob mark-and-sweep.
- Added frozen-state/CAS revalidation, fail-closed malformed JSONL handling, missing-reference integrity errors, envelope-before-blob apply ordering, per-blob re-marking, idempotent deletion, and GC deletion manifests through the shared framework.
- Wired `orbit gc audit`; converted `orbit audit prune` into a deprecated plan-first compatibility route with explicit `--apply`.
- Added focused core/store/CLI tests for shared and missing blobs, malformed JSONL, concurrent file changes, holds, retained runs, workspace isolation, restart/idempotency, legacy rows, and v2 rows.
- Updated ADR-0220, the GC design narrative, and the Unreleased changelog entry.
Validation:
- PASS: `cargo test -p orbit-core command::tests::gc_audit` (5 tests).
- PASS: focused legacy/v2 store GC tests and `cargo test -p orbit-cli command::tests::gc` (5 tests).
- PASS: `cargo clippy -p orbit-store -p orbit-core -p orbit-cli --all-targets -- -D warnings`.
- PASS: `cargo fmt --all -- --check` and `git diff --check`.
- PARTIAL ENVIRONMENT LIMITATION: `make ci-fast` reached the installer security guardrail after formatting and the pubkey check passed, but could not run `scripts/test-installer-security.sh` because `node` is not installed on PATH. No code/test failure was reported; CI remains the authoritative merge gate.
Assessment: The collector is conservative under ambiguity, never age-deletes blobs, preserves cross-workspace isolation, and is restart-safe through immutable plans, compare-and-delete store operations, file fingerprints, re-marking, and append-only deletion manifests.
Strategic decisions:
- Existing ADR-0220 already governs the relevant alternatives and safety contract, so it was updated for the audit collector rather than allocating a redundant ADR.
- Holds, exports, and retained-run bundles are durable evidence roots beneath `.orbit/state`; every SHA-256-shaped reference found there participates in the mark set.

</details>

## Validation

- Not reported

## Branch Freshness

- Head ref: `orbit/ORB-10186-6a54169e`
- Behind base: 0
- Ahead of base: 1


crates/orbit-core/src/command/gc_audit.rs
```diff
@@ -0,0 +1,592 @@
...
+            .path
+            .as_deref()
+            .ok_or_else(|| OrbitError::Execution("audit file candidate has no path".to_string()))?;
+        if !path.exists() {
+            return Ok(stale_skip("file was already removed"));
+        }
+        if file_fingerprint(path)? != candidate.expected_state {
+            return Ok(stale_skip("file changed after the plan was frozen"));
+        }
+        if let Some(hash) = candidate.id.strip_prefix("blob:")
```
danieljhkim, line 320: **[Spec compliance] — Re-marking is not atomic with blob/JSONL deletion.**

`revalidate` rebuilds the retained-reference mark set and checks the file fingerprint, then returns `Ready`; `apply` later calls `remove_file` with no lock or transaction shared by audit-event/blob/loop-JSONL writers. A concurrent writer can add a retained envelope after the re-mark and before removal, leaving that envelope pointing at a deleted blob. Likewise an actively appended JSONL partition can be unlinked between fingerprint check and write.

Introduce a workspace audit writer/GC guard honored by every v2/loop/blob publication path and by GC. Hold it across final re-mark/fingerprint validation through deletion, with lock ordering under the host GC lock. Add deterministic writer-wins and GC-wins concurrency tests; the current sequential changed-file test is not sufficient.

danieljhkim: Addressed in `ba2f9aa2` (guard commit `1fce128c`; also merged current `agent-main` through ORB-10187).

**Guard.** New `crates/orbit-common/src/utility/audit_writer_guard.rs` — a workspace-scoped advisory file lock (`<audit_root>/.gc-writer.lock`) that both the audit writer paths and the collector derive identically from the workspace `state/audit` dir, so they rendezvous on the same inode. Bounded 5s acquire; a timeout fails closed (a writer surfaces a non-fatal audit-write failure rather than publishing outside the guard; GC treats it as "do not delete").

**Writers hold it across their publication.** `V2SqliteSink` acquires the guard in `write_envelope`, `write_loop_event`, and `write_blob` — the workspace v2 event, loop JSONL append, and content-addressed blob paths (`crates/orbit-engine/src/activity_job/sqlite_sink.rs`).

**GC holds it continuously across re-mark → delete.** `AuditGcCollector::apply` now acquires the guard, re-runs the final mark/fingerprint revalidation *under* it, and only then unlinks the envelope/blob (`crates/orbit-core/src/command/gc_audit.rs`). A concurrent writer either published first — the re-mark observes the new reference / changed fingerprint and we skip (fail closed) — or it blocks until deletion completes and republishes (its blob write recreates the content-addressed object), so no retained envelope is left pointing at a swept blob and no live append is lost.

**Lock order** is fixed: host GC lock (ADR-0220) → audit writer guard → filesystem mutation. The guard is a filesystem advisory lock, never the global SQLite write lock, so no DB lock is held across a blob/JSONL deletion.

**Deterministic concurrency tests** (the sequential changed-file test is no longer the sole coverage):
- `writer_wins_published_reference_under_guard_survives_blob_sweep`
- `gc_wins_holds_guard_through_deletion_and_blocked_writer_is_not_lost`
- `active_jsonl_writer_append_under_guard_blocks_partition_sweep`
- plus `apply_re_marks_reference_added_after_planning_even_without_contention`, and `V2SqliteSink` guard-blocking tests `write_blob_blocks_on_a_held_audit_guard_then_publishes` / `write_envelope_blocks_on_a_held_audit_guard_then_persists`.

**Validation** (focused): `orbit-core gc_audit` 9/9, `orbit-engine sqlite_sink` 2/2, `orbit-cli gc` 7/7, `orbit-store audit` 16/16, `orbit-core skill_gc` 8/8 (merge-inherited); `cargo fmt --all --check` clean, `cargo clippy` (affected crates, `--tests`) clean, `git diff --check` clean.

Leaving this thread open for your verification.

danieljhkim: The shared lock closes the revalidate→unlink window for each individual write, but the blob/reference transaction is still split. `AuditSink::write_blob` acquires and releases the guard after writing the blob; the later `write_loop_event` / `write_envelope` acquires it again to publish the hash reference. GC can win between those two calls, mark the just-written blob unreferenced, delete it, and then the writer publishes an envelope pointing at a missing blob. The new tests manually publish the reference while already holding the guard, so they do not race the real production `write_blob`→event sequence and miss this interleaving. Hold one writer transaction across blob creation plus reference publication, or add an equivalent pending-publication root/recreate check, and race the actual sink API sequence against GC.

danieljhkim: Fixed in `eca678df` (merged current `agent-main` through ORB-10183/10184 first, no rebase).

**Root cause, restated.** The guard made each *individual* write atomic, but `write_blob` (guard acquired→released) and the later `write_envelope`/`emit` (guard re-acquired) are two separate critical sections. GC could sweep the just-written, not-yet-referenced blob in the gap and strand the reference. Confirmed.

**Fix — a durable per-blob pending-publication root** (`crates/orbit-common/src/utility/audit_pending.rs`). The two guarded calls are bridged by a marker that spans the whole logical transaction:

- **`write_blob`** records `state/audit/pending/<hash>` *atomically with the blob, under the same guard*, stamped with the write time (written temp+rename → crash-safe). A marker-write failure degrades exactly like a guard/blob-write failure — it returns an `error:` sentinel, so a reference to an unprotected blob is never published.
- **`write_envelope` / `write_loop_event`** clear the markers for the hashes they publish *after the row commits, under the guard* — so at no guarded instant is a blob both unmarked and unreferenced (marker → retained reference, no gap). Both the loop path (`input_sha256`/`body_sha256`) and every v2 envelope (incl. the CLI-runner `stdin/stdout/stderr_blob_ref`, which all persist through `V2SqliteSink::write_envelope`) are covered. Writer clearing and collector marking share one blob-hash tokenizer so they can't diverge.
- **`AuditGcCollector`** unions fresh markers (timestamp ≥ cutoff) into the reference set: such a blob is never a sweep candidate (`plan` skips it `pending_publication`) and `apply` fails closed on it because `blob_is_referenced` now consults the pending root under the deletion guard. A marker past the cutoff has outlived any publication window and is reclaimed as an ordinary GC candidate together with its orphan blob — so a never-published blob (e.g. the discarded `write_blob` at `agent_loop.rs:502`) leaks no longer than the retention window rather than forever, and the pending dir cannot grow unbounded.

Lock order is unchanged: host GC lock → audit writer guard → filesystem mutation.

**Tests race the *actual* sink API sequence**, not a manual publish-under-guard:

- `orbit-engine` `real_write_blob_then_write_envelope_marks_then_clears_pending` / `real_write_blob_then_emit_loop_event_clears_pending` — real `write_blob` then real `write_envelope`/`emit`; asserts the marker appears at `write_blob` (protecting the unreferenced blob) and is retired only once the reference is durable.
- `orbit-core` `fresh_pending_marker_protects_blob_and_apply_fails_closed` — a blob whose marker is fresh: `collector.apply` on a frozen sweep candidate returns `Err` (fail closed) and the blob survives — the exact interleaving from your comment.
- `orbit-core` `pending_publish_gc_lifecycle_never_strands_reference` — end-to-end mark → GC (skipped, not swept) → publish+clear → GC (referenced); proves no published reference points at a missing blob.
- `orbit-core` `stale_pending_marker_is_swept_with_its_orphaned_blob` — bounds the orphan leak.
- `orbit-common` `audit_pending` unit tests (8): mark/clear/list/malformed/tokenizer.

Existing `writer_wins…` / `gc_wins…` / `active_jsonl…` / `apply_re_marks…` and the lock order are preserved unchanged.

**Validation** (focused): `orbit-common audit_pending` 8/8, `orbit-engine sqlite_sink` 4/4, `orbit-core gc_audit` 12/12 (3 new + 9 preserved), `orbit-cli gc` 9/9, `orbit-store audit` 16/16, `orbit-core gc_logs`/`skill_gc` 8/8 each; `cargo fmt --all --check` clean, `cargo clippy -p orbit-common -p orbit-core -p orbit-engine --all-targets -D warnings` clean, `git diff --check` clean. `make ci-fast` passes fmt-check + pubkey; it can't run `test-installer-security.sh` here (`node` not on PATH) — environmental only; CI remains the merge gate.

Leaving open for your verification.

PR #585, opened 2026-07-12
Add stale managed-skill GC

danieljhkim:
## Task

[ORB-10187](https://orbit-cli.com/tasks/ORB-10187) — Add stale managed-skill GC

## Problem

After generated Orbit skills are deprecated or renamed, stale directories and links can remain across supported agent skill roots. Existing init/unlink cleanup recognizes only a subset of historical generated content.

## Why It Matters

Stale skills continue injecting obsolete instructions, but shared skill roots may also contain modified or user-authored content that Orbit must never delete by name alone.

## Scope

Implement `orbit gc skills` using the managed ownership and retirement metadata from ORB-10181.

## Constraints / Notes

- Only exact known generated hashes or proven Orbit-owned links are eligible.
- Modified/ambiguous content is reported as a conflict and retained.
- Codex/plugin caches and third-party plugin content are out of scope.

### Acceptance Criteria

- `orbit gc skills` finds retired generated skill directories and obsolete/broken Orbit-owned links across supported agent skill roots.
- Eligibility uses ORB-10181 ownership/tombstone metadata; name matching alone is never sufficient.
- Modified generated skills, same-named user content, links targeting elsewhere, plugin caches, and third-party content are retained with conflict reasons.
- Current missing/wrong link repair is reported separately from stale retirement.
- Apply removes only proven owned entries and prunes empty Orbit-owned directories without following symlinks.
- Tests cover deprecated skill, modified generated skill, user-owned same-name skill, broken owned link, escaping link, current repair, dry-run parity, and idempotency.
- `make ci-fast` and focused skill/init/CLI tests pass.

## Execution Summary

<details>

Implemented `orbit gc skills` on top of the ORB-10180 GC framework, using ORB-10181 ownership/tombstone metadata as the sole eligibility proof.

New: crates/orbit-core/src/command/skill_gc.rs — SkillsGcCollector (impls GcCollector). plan() scans the global generated-skill root (~/.orbit/skills) and per-agent link roots (~/.agents/skills, ~/.claude/skills, resolved via a new pub(crate) init::skill_gc_roots helper that shares init's link-root logic). Two removal classes: (1) retired generated directories — classify_install==OrbitOwned for a *tombstoned* id (fingerprint-proven; name matching alone insufficient); (2) stale/broken owned links — classify_symlink==OrbitOwned whose id is tombstoned or whose target is broken. Retained as conflict skips: modified generated (Ambiguous->modified_generated), same-named user dirs (Unmanaged->user_content), foreign links (Ambiguous->foreign_link), non-symlink entries (user_content). Current-skill missing/wrong links are reported separately as link_repair skips (never deleted).

Mutation-safety split: directory-delete candidates carry path=Some so the framework's non-bypassable validate_candidate_path gate guards remove_dir_all (a matching fingerprint proves the tree holds no symlinks); link candidates carry path=None (agent roots are siblings of the scope root) and the collector self-validates (symlink_metadata + re-classify under the GC lock) before a no-follow remove_file, then prunes emptied owned link dirs with remove_dir. revalidate re-derives ownership from a fresh manifest load so a concurrent edit can't slip an ambiguous entry through.

CLI: crates/orbit-cli/src/command/gc.rs — `gc skills` now dispatches to SkillsGcCollector (global scope from collector.scope()); refactored shared run/finish helpers off execute. Other targets keep EmptyGcCollector.

Docs: CHANGELOG Unreleased bullet; docs/design/gc/2_design.md ORB-10187 task ref marked implemented (ADR-0220 already covers the contract; no new ADR).

Tests (crates/orbit-core/src/command/tests/skill_gc.rs, 6): deprecated skill removal, modified generated retained, user same-name retained, broken owned link removal, escaping/foreign link retained (target never traversed), current-skill link repair reported separately, dry-run/apply parity, idempotency, empty-owned-dir pruning. All pass.

Validation: cargo fmt --all --check clean; clippy -p orbit-core -p orbit-cli --all-targets clean; skill_gc (6), skill_ownership (18), init::tests (5), cli command::tests::gc (4) pass; check-changelog-style.sh ok.

Not run in this sandbox: full `make ci-fast` — its fmt-check passes but scripts/test-installer-security.sh aborts with 'required binary node not on PATH' (environment limitation; CI runs it unrestricted). Unrelated pre-existing failure: cli tests::audit_middleware::tool_run_audit_meta_uses_agent_role_without_identity fails only because this executor sets ORBIT_AGENT_MODEL=opus (the run's crew); with that env unset it passes — not caused by this change.

Field note (not a code issue): the live ~/.orbit/skills/.ownership.json is from an intermediate unreleased ORB-10181 build (old field name, missing tree_fingerprint), so `gc skills` fails closed there with a corrupt-manifest error rather than deleting — correct behavior; the stale manifest just needs regenerating via init. No user impact (ORB-10181 unreleased).

</details>

## Validation

- Not reported

## Branch Freshness

- Head ref: `orbit/ORB-10187-6a541ab3`
- Behind base: 0
- Ahead of base: 1


crates/orbit-core/src/command/skill_gc.rs
```diff
@@ -0,0 +1,574 @@
...
+                        code: code.to_string(),
+                        reason,
+                    }),
+                }
+            }
+        }
+
+        // 2. Stale/broken owned links + current-skill repair reporting.
+        for link_root in &self.link_roots {
+            if !link_root.is_dir() {
```
danieljhkim: [P1] Report missing current links when an entire managed root is absent

This early continue skips the repair-reporting loop below, so if ~/.agents/skills or ~/.claude/skills does not exist, orbit gc skills reports no link_repair entries at all for current managed skills. That directly misses the acceptance case for current missing links and makes a fully absent root look healthy. Treat a missing root as an empty present set and emit link_repair for every current skill (while retaining non-directory/ambiguous roots fail-closed), then add coverage for the absent-root case.

crates/orbit-cli/src/command/gc.rs
```diff
@@ -66,25 +68,49 @@ pub struct GcCommand {
 impl Execute for GcCommand {
     fn execute(self, runtime: &OrbitRuntime) -> Result<(), OrbitError> {
         let target = GcTarget::from(self.target);
-        let defaults_global = matches!(target, GcTarget::Logs | GcTarget::Skills);
+        // Skills is a global-only collector whose owned state spans the global
+        // generated-skill root and the per-agent link roots; it resolves its own
+        // scope rather than following the generic workspace/global split.
+        if target == GcTarget::Skills {
```
danieljhkim, line 92: [P1] Reject workspace scoping instead of silently doing global GC

For the global-only skills target this branch ignores both --workspace and the selected scope, then proceeds against the global generated/link roots. A caller who explicitly passes --workspace can therefore get a global apply despite asking for workspace scope. Fail with InvalidInput when --workspace is present (and test that no planning or mutation occurs); --global or the default can continue to select this collector.

crates/orbit-core/src/command/skill_gc.rs
danieljhkim: Fixed in 0df8c47e (merge of current agent-main + this fix). The link-root loop now matches on `fs::symlink_metadata(link_root)`:

- **Absent root** (`NotFound`) → empty `present` set, so the repair-reporting loop below emits a `link_repair` skip for **every** current managed skill (a fully-absent root no longer reads as healthy).
- **Real directory** → scanned as before.
- **Exists but not a directory** (a file, or a symlink we must not follow) → retained fail-closed as a `foreign_root` skip; it is neither traversed nor repair-reported through.

Coverage added in `crates/orbit-core/src/command/tests/skill_gc.rs`: `absent_managed_root_reports_repair_for_every_current_skill` (asserts link_repair for each current skill under both absent roots, no removals, and no mutation on apply) and `non_directory_managed_root_is_retained_fail_closed`.

crates/orbit-cli/src/command/gc.rs
danieljhkim: Fixed in 0df8c47e. `GcCommand::execute` now rejects an explicit `--workspace` on the global-only skills target with `OrbitError::InvalidInput` **before** constructing `SkillsGcCollector` or planning/mutating anything:

```rust
if target == GcTarget::Skills {
    if self.workspace.is_some() {
        return Err(OrbitError::InvalidInput(
            "`gc skills` operates on global Orbit-owned state; the `--workspace` selector is not supported".to_string(),
        ));
    }
    let collector = SkillsGcCollector::for_global_root(&runtime.paths().global_dir);
    ...
}
```

The default and explicit `--global` still select the skills collector unchanged. Coverage: `gc_skills_rejects_workspace_selection_without_mutation` in `crates/orbit-cli/src/command/tests/gc.rs` seeds+retires a generated skill (a genuine removal candidate) on an in-memory runtime, runs `gc skills --workspace ws_example --apply`, and asserts the error is `InvalidInput` and that the retired directory a global apply would have reclaimed is still on disk (no planning, no mutation).

PR #586, opened 2026-07-12
Add terminal job-run bundle and artifact retention to orbit gc

danieljhkim:
## Task

[ORB-10183](https://orbit-cli.com/tasks/ORB-10183) — Add terminal job-run bundle and artifact retention to orbit gc

## Problem

Terminal job runs span authoritative SQLite rows, step records, legacy/regenerable run bundles, scoreboards/checkpoints, task references, and v2 audit evidence. Existing archive/delete operations are one-run-at-a-time and do not provide a retention collector.

## Why It Matters

Run artifacts grow continuously, but naive directory deletion can strand metadata or destroy evidence needed to diagnose retained failures and audit pipeline behavior.

## Scope

Implement `orbit gc runs` with separate archive and purge ages, coordinated deletion across run-owned surfaces, and explicit boundaries with the audit collector.

## Constraints / Notes

- Active, resumable, or liveness-inconclusive runs are never eligible.
- Eligibility uses terminal transition time, not filesystem mtime.
- Audit envelopes/blobs are governed by `orbit gc audit`, not removed incidentally.

### Acceptance Criteria

- `orbit gc runs` inventories every persisted surface owned by a job run and reports coordinated eligibility before mutation.
- Only conclusively terminal runs are eligible; pending/running/resumable/liveness-inconclusive runs are retained.
- Separate configurable archive and purge ages preserve recent failed/interrupted runs longer for diagnosis.
- Eligibility uses the persisted terminal transition timestamp.
- Apply reuses `archive_job_run`/`delete_job_run` primitives where safe and cannot strand SQLite rows, steps, reservations, scoreboards/checkpoints, task references, or legacy bundles.
- Audit JSONL/envelopes/blobs are excluded from run deletion and delegated explicitly to `orbit gc audit`.
- Tests cover active, failed-retained, terminal-expired, stale legacy bundle, task-linked, partial-failure, and idempotent cases.
- `make ci-fast` and focused store/core/CLI tests pass.

## Execution Summary

<details>

Outcome: success
Changes:
- Added `RunGcCollector` and wired `orbit gc runs` with plan/apply reporting, per-run claim-guard revalidation, separate success/failure archive and purge ages, terminal-transition clocks, and explicit audit exclusion.
- Added SQLite schema migration v4 (`job_runs.archived_at`). Archive hides rows from ordinary queries while preserving steps/checkpoints; purge transactionally deletes released owner reservations, invocation/scoreboard inputs, the run row, and cascading steps. Active reservations, task/retry links, persisted recovery checkpoints, aggregate scoreboard references, and live/inconclusive owners are retained.
- Added restart-safe legacy bundle archive/purge handling, including stale rowless bundles and idempotent recovery from partial filesystem/store mutations.
- Added config/CLI/store/core coverage for active, failed-retained, resumable, live-owned, terminal-expired, task-linked, stale legacy, partial-failure, migration, cascade, and idempotent cases; updated GC design docs and CHANGELOG.
Assessment: The collector fails closed on every ambiguous ownership/liveness/reference condition, uses the persisted terminal timestamp exclusively, and leaves audit JSONL/envelopes/blobs untouched for `orbit gc audit`. Aggregate scoreboard files are treated as holds rather than rewritten, preserving evidence without dangling references.
Validation:
- `cargo test -p orbit-store sqlite::migration::tests` (14 passed), `sqlite::job_run_store` (6 passed), `cargo test -p orbit-core command::tests::gc` (7 passed), and `cargo test -p orbit-cli command::tests::gc` (6 passed).
- Targeted run-GC/config tests passed after live-owner and resumable coverage was added.
- `cargo check -p orbit-cli`, `cargo fmt --all -- --check`, `git diff --check`, and `cargo clippy -p orbit-store -p orbit-core -p orbit-cli --all-targets -- -D warnings` passed.
- `make ci-fast` reached `scripts/test-installer-security.sh` and could not run it because this runner has no `node` binary. All remaining ci-fast guardrails were run directly and passed; `check-installer-pubkey.sh` also passed.
Tooling note: required Orbit CLI search/graph/task/friction calls were blocked first by stale crew-schema parsing and then by a stale skill ownership manifest missing `tree_fingerprint`; plan, scope comments, and this summary were persisted through the dashboard PATCH surface. The friction report command was attempted and failed for the same manifest validation error.

</details>

## Validation

- Not reported

## Branch Freshness

- Head ref: `orbit/ORB-10183-6a542275`
- Behind base: 0
- Ahead of base: 1


crates/orbit-core/src/command/gc/runs.rs
```diff
@@ -0,0 +1,653 @@
...
+                                message: error.to_string(),
+                            });
+                            continue;
+                        }
+                    };
+                    if persisted_ids.contains(&run.run_id) {
+                        continue;
+                    }
+                    plan.scanned = plan.scanned.saturating_add(1);
+                    let Some(finished_at) = run.finished_at.filter(|_| run.state.is_terminal())
```
danieljhkim, line 527: [P1] Apply the same protection contract to rowless legacy bundles

This rowless path declares a bundle eligible from terminal state and age alone. It skips the persisted-run protections above: owner liveness/PID identity, resumable checkpoint state, task and retry links, active reservations, and scoreboard references. Apply then mutates it without the shared per-run claim guard and without rechecking that an authoritative row has appeared. A terminal legacy bundle can therefore be archived/purged while a live or resumable owner still uses it, or while retained evidence still references it, violating criteria 2 and 5. Factor the common protections across persisted and legacy candidates, hold the per-run guard through final no-row/protection revalidation and mutation, and add rowless legacy tests for live owner, resumable checkpoint, retained task/reference, plus row-created-between-plan-and-apply.

crates/orbit-cli/src/command/gc.rs
```diff
@@ -82,6 +98,15 @@ impl Execute for GcCommand {
...
+            || self.purge_after_days.is_some()
+            || self.failure_archive_after_days.is_some()
+            || self.failure_purge_after_days.is_some();
+        if target != GcTarget::Runs && has_run_override {
+            return Err(OrbitError::InvalidInput(
+                "run retention overrides require the `runs` target".to_string(),
+            ));
+        }
         let defaults_global = matches!(target, GcTarget::Logs | GcTarget::Skills);
         let scope = if self.global || (self.workspace.is_none() && defaults_global) {
```
danieljhkim, line 144: [P1] Reject the global scope for this workspace-only collector

The normative contract marks Runs as workspace-only and says unsupported scope/target combinations must be not_applicable. Here orbit gc runs --global constructs an OrbitRuntime rooted at the global Orbit directory and runs the collector there, potentially creating/scanning the wrong state tree instead of refusing the scope. Reject --global for runs (and worktrees if not already handled by its collector), before constructing a runtime or planning, and add CLI coverage proving no mutation occurs.

crates/orbit-core/src/command/gc/runs.rs
danieljhkim: Fixed in 128fa5bc — rowless legacy bundles now enforce the **same** protection contract as authoritative rows.

**Shared protections.** Factored the ownership/liveness/reference holds into `RunGcCollector::protection_hold` — owner liveness/PID identity (`gc_owner_permits_reclaim`), resumable recovery checkpoints (`is_resumable`), retained task links, active reservations, retry-source links, and aggregate scoreboard references. Both `classify_record` (persisted rows) and `inventory_stale_legacy` (rowless bundles) now call it, so a legacy bundle is never eligible from terminal state + age alone.

**Guarded apply + no-row revalidation.** The legacy `apply` branch now acquires the shared per-run claim guard (`run_claim_guard::acquire` on `state/run-guards/`, the same lock the claim/start path holds) and, while holding it, re-runs `revalidate_stale_bundle`, which additionally **fails closed if an authoritative row has appeared** (`row_appeared` → the persisted collector owns it) and re-runs `protection_hold` on the freshly-read bundle before any archive/purge. So a live/resumable owner, a retained reference, or a row materializing between plan and apply cannot race the mutation.

**Tests** (`crates/orbit-core/src/command/tests/gc.rs`):
- `run_gc_rowless_legacy_bundle_fails_closed_on_live_owner_task_and_retry_references` — rowless bundles held on `live_or_inconclusive`, `task_linked`, `retry_linked`; a clean rowless bundle stays eligible (proving the protections gate rather than blanket-skip).
- `run_gc_rowless_legacy_bundle_skips_and_holds_guard_when_row_appears_before_apply` — a row (live owner + resumable checkpoint) reappears after the plan froze; `revalidate` returns `row_appeared`, and guarded `apply` refuses to mutate — the bundle and the reappeared row both survive.

Note on resumable-rowless: a persisted checkpoint is row-resident (`pipeline_state_json` on `job_runs`), so a resumable run implies a row exists — that path is caught by `row_appeared` / the persisted collector; `protection_hold` still wires `is_resumable` for symmetry, and the row-appeared test plants exactly that (interrupted + checkpointed live row).

`cargo test -p orbit-core command::tests::gc` (18 passed), `fmt --check`, `clippy -p orbit-store -p orbit-core -p orbit-cli --all-targets -D warnings`, and `git diff --check` all pass.

crates/orbit-cli/src/command/gc.rs
danieljhkim: Fixed in 128fa5bc. `GcCommand::execute` now rejects `orbit gc runs --global` up front:

```rust
if self.global && target == GcTarget::Runs {
    return Err(OrbitError::InvalidInput(
        "run garbage collection is workspace-only; `--global` is not a supported scope".to_string(),
    ));
}
```

This runs **before** resolving scope, constructing the collector `OrbitRuntime` (`from_roots`), acquiring the host GC lock, or planning — so the workspace-only `runs` collector (design §3.2) refuses rather than scanning/mutating the global state tree.

**Test** (`crates/orbit-cli/src/command/tests/gc.rs`): `gc_runs_rejects_global_scope_before_any_mutation` invokes `runs --global --apply`, asserts `InvalidInput`, and asserts no GC manifest/lock state is created under the global root (`state/gc` absent) — proving no mutation.

Scoped to `runs` per this task. The `worktrees` collector is owned separately (ORB-10182); its scope handling is untouched here.

PR #684, opened 2026-07-25
fix: Carry invocation telemetry columns to existing stores; never fail a run on a telemetry write [ORB-10367]

danieljhkim:
Fixes **ORB-10367** / friction **F2026-07-105** — the incident that killed every agent-dispatching pipeline run on dk-server-1 tonight.

## Root cause: migration exists, but never runs on existing databases

Not a missing migration, and not a wrong database path.

`ensure_invocation_schema` holds the `ALTER TABLE invocations ADD COLUMN` steps for `cache_create_1h_tokens` and `provider_cost_usd`, and it is only called from `apply_baseline_schema` — migration **v1**. `run_migrations` applies entries with `version > current`, so any database that already recorded v1 skips baseline forever. Both columns were added to baseline on 2026-07-20; `~/.orbit/orbit.db` recorded `migration.v0001` on 2026-07-04. The ALTERs never reached it, the insert bound columns the table lacked, and the run died at the telemetry write — after the agent work had finished.

Distinguishing evidence: `provider_cost_usd` (same function, added the same day) is **missing from every live database on the box** while the insert binds it. Nothing has ever applied that ALTER anywhere. Path is not the variable — `audit_db` is `global_dir/orbit.db` and `Store::open` migrates the very handle it writes through, so the migrated database and the written database are the same file by construction.

## Changes

**1. Schema (`orbit-store`)** — register the idempotent invocation-schema step as migration **v10** `invocation_telemetry_columns`, guarded on the tables existing. `SUPPORTED_SCHEMA_VERSION` → 10. Both drifted columns are fixed together, not just the one the error named.

**2. Telemetry is non-fatal (`orbit-engine`)** — the durable half. `persist_dispatch_invocation` no longer returns a `Result`: there is no error path left to propagate into a step outcome. Both engine call sites (dispatcher path and HTTP agent-loop path) record the failure through the new `V2AuditWriter::note_telemetry_failure`, mirroring the ORB-00414 audit-failure pattern:
- `tracing::error!` on `orbit.engine.telemetry` naming run, step, and error;
- a `telemetry.persist_failed` audit event on the run record;
- `JobOutcome::{telemetry_failures, degraded_telemetry}`.

A run's success stays decided solely by its work. Had this been in place, four runs' completed work would not have been thrown away.

**3. Structural regression test** — the insert is rendered from one `INVOCATION_INSERT_COLUMNS` list. The new test stamps a database at the newest pre-fix ledger version with the pre-telemetry `invocations` shape (the live dk-server-1 shape), migrates it, and asserts every insert-bound column exists *and* that a real `insert_invocation_trace_record` lands. Verified to fail against the broken state with the exact production message: \`migrated database is missing insert-bound column \`cache_create_1h_tokens\`\`. A fresh-database test cannot catch this class.

## Deployment notes

- **No manual database step is needed** once this ships: v10 applies on the next `Store::open`. Restart `orbit-web.service` after the binary swap.
- **Sequencing vs ORB-10361 (redeploy):** a redeploy alone does *not* fix this and makes it worse — the current live binary already binds `provider_cost_usd`, which is missing from every database on the box, so redeploying without this change just moves the fatal error to the next column. Land this first.
- **Downgrade guard:** databases migrated to v10 are refused by older orbit binaries, as with every prior schema bump.

## Validation

`cargo fmt`, `cargo clippy --workspace --all-targets -D warnings`, workspace tests, doc tests, `cargo doc -D warnings`, `cargo deny`, and every guardrail script pass. Two failures on the box are environmental and reproduce on a pristine `agent-main` checkout: `direct_search_semantic_command_surfaces_companion_stderr` (verified against stashed baseline) and `git_commit_batch_uses_templated_single_task_message` (fails only because the worker run exports `AGENT_RUN_ID`/`AGENT_MODEL`/`AGENT_TASK`, which the commit-trailer feature picks up; 285/285 pass with those unset).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- **Telemetry no longer discards completed runs**: the `invocations` insert-bound columns (`cache_create_1h_tokens`, `provider_cost_usd`) reach existing databases through a new `invocation_telemetry_columns` migration (schema v10) instead of only the v1 baseline, and a failed invocation-trace write is now logged and recorded as a `telemetry.persist_failed` event on the run rather than failing the job. ([ORB-10367])
 - **PR review status clearing is documented**: `orbit.task.update` now describes its empty-string clear convention, with tool-host coverage for the persisted null result. ([ORB-10229])
 - **Task PR status updates persist in v2 storage**: local tool and dashboard PATCH updates now retain `pr_status` alongside status and execution-summary changes. ([ORB-10223])
 - **Bundled skills are repository-agnostic**: the embedded skill tree drops Orbit-source paths, private Constellation names, workspace-local artifact IDs, and fixed design-doc filenames; a portability regression test and byte-aligned plugin mirrors guard against reintroduction. ([ORB-10208])
--- a/crates/orbit-common/src/types/activity_job/audit_envelope.rs
+++ b/crates/orbit-common/src/types/activity_job/audit_envelope.rs
@@ -184,6 +184,16 @@ pub enum V2AuditEventKind {
         harness_version: Option<String>,
         timed_out: bool,
     },
+    /// [ORB-10367] A telemetry write failed (e.g. persisting an invocation
+    /// trace). The run continues — telemetry is observability, not
+    /// correctness — but the failure is recorded on the run so the gap in
+    /// the telemetry is visible rather than silent.
+    TelemetryPersistFailed {
+        component: String,
+        #[serde(default, skip_serializing_if = "Option::is_none")]
+        step_id: Option<String>,
+        error: String,
+    },
 }
 
 impl V2AuditEventKind {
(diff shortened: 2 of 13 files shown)


danieljhkim:
**CI status: the one red test is a pre-existing `agent-main` failure, not this PR.**

`Check / Clippy / Test` — 1634 of 1635 executed tests pass; the single failure is
`orbit-cli::semantic_companion direct_search_semantic_command_surfaces_companion_stderr`
([run 30141527687](https://github.com/danieljhkim/orbit/actions/runs/30141527687)). nextest stops the
run there, so 941 tests after it never executed.

Evidence it is not this change:
- The last five `agent-main` CI runs are all `failure` on the same test — including
  [run 30140579236](https://github.com/danieljhkim/orbit/actions/runs/30140579236) at commit
  `66f848a3`, the base of this branch.
- It also fails on a pristine `agent-main` checkout locally (verified by stashing this diff and
  re-running the suite).
- It touches search-companion fallback behavior; this PR touches store migrations and the engine's
  invocation-trace persistence.

Both tests added here ran and passed in CI:
- `orbit-store sqlite::migration::tests::ledger::migrated_legacy_db_carries_every_invocation_insert_column`
- `orbit-engine activity_job::job_executor::tests::target::failed_invocation_trace_persist_does_not_fail_the_step`

Filed as friction F2026-07-107 — `agent-main` being red for every PR is worth its own task.

PR #689, opened 2026-07-25
feat(store): Resolve crew aliases to exact model strings before the invocation insert [ORB-10354]

danieljhkim:
Resolves **ORB-10354**. Implements **ADR-0247** (included in this diff for review).

## Problem

Orbit's crew config dispatches Claude through unversioned CLI aliases (`opus`, `sonnet`, `fable`) and gemini through `pro`, so CLI defaults never drift. Those alias strings flowed straight into `invocations.model`, which ADR-0245 keys by **exact** model string. No price row could ever match an alias, so alias-dispatched runs derived no cost and sat in per-model aggregates as if each alias were its own model.

## Change — resolve at ingest, in the store

**`orbit_common::types::model_identity`** (new)
- One alias table mapping `(agent family, alias)` → the exact string that alias currently resolves to, plus a classifier returning `Exact` / `ResolvedAlias` / `UnresolvedAlias`.
- gemini `pro` is deliberately left unresolved — Orbit pins two conflicting exact gemini pro strings, so a guess would fabricate provenance.

**Store — `invocations.model_alias`**
- `insert_invocation_trace_record` splits the caller's model string across two columns: `model` always carries an exact provider string (or `NULL`), `model_alias` carries the alias the invocation was dispatched through.
- Normalization lives in the store because that call is the one funnel every ingest path shares — including the worker bridge's `POST /api/metrics/invocations`, which deserializes `InvocationInsertParams` straight from JSON and never passes through the runtime's identity normalization.
- **Migration v11 `invocation_model_alias`** adds the column and lifts historical alias values out of `model` into `model_alias` *without* resolving them (an `opus` row from June may not mean today's flagship). Registered as its own migration step rather than folded into an earlier one. `SUPPORTED_SCHEMA_VERSION` 10 → 11.

**Price coverage**
- New `Store::list_unpriced_invocation_models` live coverage scan — possible only because aliases left the `model` column. Aliases are deliberately unpriced, so before this change every scan would have reported them as permanent false positives. The curated unit-level fleet list stays as the CI-runnable form, since CI has no invocation ledger to scan.

**Docs** — `docs/design/auditability/2_design.md` and `4_decisions.md` updated in-PR; ADR-0247 accepted.

The alias-staleness trade-off and the deferred authoritative fix (read the model the provider reports in its result JSON) are recorded in ADR-0247 and tracked as **ORB-10370**.

## Validation

From the original run's execution summary: `make ci-fast` pass, `cargo clippy --workspace --all-targets` clean, `cargo test` green across orbit-common / orbit-store / orbit-core / orbit-engine / orbit-dashboard, **13 new tests**. CI on this PR is the authoritative gate.

## Provenance — recovered by a rescue run

Job run `jrun-20260725-0457-3` (claude/opus) **completed this implementation** and then died on the `worktree_integrity_ambiguous` boundary check: the primary checkout advanced mid-run (`b63622c0` → `ce8e7053`, the PR #687 merge plus an unrelated learning-CLI merge). Every path in the guard's `changed_paths` belonged to that unrelated merge; this run touched none of them. Nothing was committed or pushed, so the finished work was stranded in the worktree.

This PR is a hand-dispatched rescue that recovered it **as-is** — no re-implementation, no "improvements". All 7 new files were verified against their expected blob hashes before committing, and the rebase onto `agent-main` was verified to leave the commit's patch byte-identical.

- Rescue agent-run: `715721ce-9963-48a4-9d93-b119153be88f`
- Rebase onto `origin/agent-main` (`bb94f032`): **clean, zero path overlap** with the 3 intervening base commits.
- This should be the last rescue of this shape — PR #688 (ORB-10363) narrowed the guard and added an automatic PR-failure handoff. This run predates that binary.

## Note for review — one artifact was quarantined

The executor also authored a **learning artifact `L-0111`**. Standing constellation policy is that executors create frictions only; learnings are human/orchestrator-authored. It was **not** committed and **not** deleted — it is preserved verbatim at `~/.orbit/quarantine/2026-07-25-ORB-10354-executor-authored-learning/` (blob `e4b1b16103d3e0c8fcf44d31d62deee8966069c2`) with a README explaining the disposition. Its content — how bumping `SUPPORTED_SCHEMA_VERSION` breaks four ledger-test sites that enumerate the migration registry — looks like genuinely reusable knowledge and may be worth authoring properly. No Orbit learning-store command was run and no other learning was touched.

**ADR-0247 is deliberately left in this diff** so it gets reviewed here.


danieljhkim:
Closing without merging — the fix moved to a different layer than this PR implements.

The root cause here was correctly diagnosed: invocation rows were landing keyed by crew *alias* (`opus`, `sonnet`) instead of an exact model string, polluting per-model cost aggregates. This PR fixes it inside `orbit-common`/`orbit-store`: a new alias-resolution table, `classify_model_string()`, a `model_alias` column, and schema migration v11.

Turns out the resolution mechanism this needed already exists and already works — the `crew_model` field on task/run records, driven by each crew's `model` value in the relevant `config.toml`. It already resolves correctly for the gpt crews (`sol` → `gpt-5.6-sol`, `terra` → `gpt-5.6-terra`); the Claude crews (`opus`, `sonnet`, `fable`) were simply configured with the alias pointing at itself. That's the entire defect, and it's a one-line-per-crew config change, not a code change.

Two reasons the configuration fix is preferred over this PR's approach:
- A table of "today's flagship" models compiled into Rust goes stale on every provider model release, turning a config edit into a code change plus a release — precisely what the `crew_model` field/config indirection exists to avoid.
- The v11 schema migration is a permanent commitment (every future binary carries it forward) that the configuration fix doesn't require at all.

None of this reflects on the implementation itself — it's careful, well-tested work (13 new tests, clean clippy, full validation pass). It solves the problem at the wrong layer, which is a design call made after the fact, not a defect in the execution.

ADR-0247 stays with this PR/branch as the record of the rejected approach — not being relocated or re-filed.

Tracking: ORB-10354.

PR #1048, opened 2026-08-16
Make dashboard failure metrics incident-aware instead of counting repeated events as independent failures

danieljhkim:
## Task

[ORB-10871](https://orbit-cli.com/tasks/ORB-10871) — Make dashboard failure metrics incident-aware instead of counting repeated events as independent failures

## Problem
The dashboard's scorecard flattens raw failed audit events into an agent reliability number. In the live Orbit view on 2026-08-15, Codex showed 446 failed calls out of 1,113, but the Audit drill-down showed that 422 of those failures were one repeated `fs.delete` pattern clustered at roughly the same time. The UI therefore presented one burst as hundreds of apparently independent quality failures.

The Errors view also renders propagation layers such as `implement_one`, `implement_bundle`, `require_child_success`, and `require_gate_success` as separate errors even when they originate from one failed run.

## Why It Matters
Raw event volume is useful for forensics, but treating it as incident count produces misleading scorecards and poor operational decisions. Operators need both the exact raw evidence and a root-cause-oriented view.

## Constraints / Notes
- Preserve raw event counts and access to every underlying audit record.
- Add a deterministic, project-agnostic incident/grouping contract based on durable metadata and error signatures; do not hardcode `fs.delete`, Codex, a workspace, or the observed counts.
- Do not silently discard expected negative-path, policy-denial, or propagated failure evidence. Classify and label it.
- Related completed work: ORB-10588 (pipeline reliability surface).

### Acceptance Criteria

- Scoreboard and diagnostics distinguish raw failed events from grouped failure incidents, and both counts are visibly labeled with their denominators and selected window.
- A synthetic burst of hundreds of identical failures is shown as one grouped incident with its raw event count, while unrelated failures remain separate incidents.
- Cascading pipeline errors from one run expose one root failure with the downstream propagation chain collapsed beneath it rather than implying independent root causes.
- Clicking an incident reveals the exact underlying audit events, actor, tool, run/task identifiers when present, first/last timestamps, and grouping signature.
- Expected negative-path failures, denials, and genuine unexpected failures remain distinguishable; no evidence is dropped from raw Audit views or exports.
- Grouping is project-agnostic and covered by deterministic store/API/UI tests with no hardcoded observed actor, tool, workspace, task, or event IDs.
- Rendered validation at desktop and a 480-720px viewport confirms the incident summary and expansion remain scannable without clipping, and focused tests plus git diff --check pass.

## Execution Summary

<details>

Made dashboard failure metrics incident-aware while keeping every raw audit row intact.

## Grouping contract (orbit-store)
New `crates/orbit-store/src/sqlite/audit_event_store/incident.rs` owns the whole contract, reading only durable audit columns:
- **Classification** — `denied` (status, or a `policy denied` / `capability denied` / `permission denied` message), `expected` (an `OrbitError`-derived caller-input negative path: invalid input, not found, validation failed, ...), otherwise `unexpected`. Classes never merge into one incident, so denials and expected negative paths stay separable from genuine failures.
- **Signature** — `class | role | surface | normalized message`, where `normalize_message` replaces volatile tokens by *shape* (`<path>`, `<num>`, `<id>`, `<uuid>`, `<hex>`, `<ts>`); a message-less failure falls back to its exit code. Nothing names a tool, agent, workspace, or task.
- **Clustering** — by `(job run, signature)`, so a burst of hundreds of identical refusals is one cluster carrying its raw `event_count` and a bounded sample of the exact rows.
- **Cascade collapse** — within one run, same-class clusters within `CASCADE_WINDOW_SECS` (60s) of each other collapse onto the earliest cluster (the root); the rest become its ordered `propagation` chain. A failure later in the same run, beyond that window, stays its own incident.
Output is deterministic and order-independent; `incident_id` is an FNV-1a hex handle over the grouping key.

`Store::get_failure_incidents` queries only `status != 'success'` rows and reports `raw_failed_events`, per-class raw/incident counts, and a `truncated` flag instead of silently capping. Added to `AuditEventStoreBackend` + the SQLite backend, exposed as `OrbitRuntime::audit_failure_incidents`. Refactored the three whole-row audit queries onto one shared `AUDIT_EVENT_COLUMNS` const rather than adding a fourth copy.

## API (orbit-web)
- New `api/incidents.rs` + `GET /api/audit/incidents?since=&class=&role=&limit=`: incidents with signature, class + label, actor, surface, first/last timestamps, run/task ids, propagation chain, and sample raw events — alongside the denominators (`total_events`, `raw_failed_events`, `incident_count`, per-class maps) and the echoed `window`. `since=all` is a valid lifetime scope; an unknown class is a 400. Messages and signatures pass through `redact_all`.
- `/api/audit/summary` gained `failed_events`, `failure_incidents`, and per-class breakdowns over the same window.
- `/api/scoreboard` gained per-agent `failure_incidents`, `failure_incident_events`, `unexpected_failure_incidents`, rolled up by canonical agent family (same normalization the existing audit overlays use) and honoring the selected window. `failed_tool_calls` / `tool_calls` are untouched.

## Dashboard
- Diagnostics gains an **Incidents** subtab (routed, aggregate-guarded, fetched against the shared dashboard window). Header reads `N incidents / M failed events`; the summary reads `grouped from M failed events of T audited events · window <w>`, with per-class chips showing `incidents/raw events`. A truncated scan says so.
- An incident expands to its grouping signature, classification, actor, surface, step, first/last timestamps, run/task ids, the propagation chain (labeled "downstream failures, not independent root causes"), and a table of the exact underlying audit events — plus buttons into the raw Audit view (now carrying the surface as a `tool` filter) and the owning run.
- Scoreboard keeps `tool fail/all` and adds a `fail inc/events` pair; every section badge names the selected window.
- CSS adds incident styling with 720px/520px reflow so the summary and expansion stay scannable without clipping.

## Validation
- `cargo test -p orbit-store --lib` 284 passed (11 new grouping tests: burst collapse, unrelated failures stay separate, cascade root + chain, post-window failure stays independent, three classes never merge, late-translated denial, order independence, normalization, exit-code fallback, store query preserves the raw view, truncation reported).
- `cargo test -p orbit-web --lib` 234 passed (6 new incidents endpoint tests, plus new audit-summary and scoreboard tests and a new `dashboard_assets` test pinning the UI contract). `cargo test -p orbit-core --lib` 795 passed.
- `make ci-fast` and `make ci-lint` both pass; `git diff --check` clean.
- Rendered validation: no browser engine is available in this runner, so pixel-level checking was not possible. Instead the real renderer was driven offline against a fixture payload through a minimal DOM shim — collapsed and expanded incident trees render correctly with all required fields and no errors — and the 480–720px reflow rules are asserted against `dashboard.css` in `dashboard_assets.rs`, matching the repo's existing convention for this class of AC (see `dashboard_managed_execution_cost_panel_has_responsive_presentation_hooks`). Unrestricted CI can arbitrate anything beyond that.

## Out-of-scope files touched
`crates/orbit-store/src/backend/{contracts/traits.rs,sqlite_backends.rs}` (the backend trait the runtime dispatches through — required to reach the new store query), `crates/orbit-core/src/{lib.rs,command/audit_event.rs}` (runtime method + re-exports), `crates/orbit-web/src/api/mod.rs` (route + module), and `crates/orbit-web/src/api/tests/reliability.rs` (pre-existing rustfmt violation in the base that `make ci-fast` rejects; formatting-only).

</details>

## Validation

- Not reported

## Branch Freshness

- Head ref: `orbit/ORB-10871-6a815a05`
- Behind base: 0
- Ahead of base: 1

### DIFF
--- a/crates/orbit-core/src/command/audit_event.rs
+++ b/crates/orbit-core/src/command/audit_event.rs
@@ -5,6 +5,7 @@ use orbit_common::types::{
 use orbit_store::{
     AuditEventFilter, AuditEventInsertParams, AuditRoleAggregate, AuditToolAggregate,
     AuditToolCallCountsByRole, AuditToolCallCountsBySurfaceAndRole, AuditTopToolCall,
+    FailureIncidentQuery, FailureIncidentReport,
 };
 
 use crate::OrbitRuntime;
@@ -243,6 +244,19 @@ impl OrbitRuntime {
         self.stores()
             .audit_events()
             .get_audit_event_aggregates_by_role(since)
+    }
+
+    /// Failure incidents grouped from the raw failed/denied audit rows in
+    /// `query`'s window [ORB-10871]. The report carries both the incident
+    /// list and the raw event denominators it was derived from, so a caller
+    /// never has to infer one count from the other. Raw rows are neither
+    /// mutated nor withheld — `list_audit_events_filtered` remains the
+    /// authority on evidence.
+    pub fn audit_failure_incidents(
+        &self,
+        query: &FailureIncidentQuery,
+    ) -> Result<FailureIncidentReport, OrbitError> {
+        self.stores().audit_events().get_failure_incidents(query)
     }
 
     /// Sorted `duration_ms` values for audit events with NULL `tool_name`
--- a/crates/orbit-core/src/lib.rs
+++ b/crates/orbit-core/src/lib.rs
@@ -86,6 +86,12 @@ pub use orbit_store::{
     AuditEventFilter, AuditEventInsertParams, AuditToolAggregate, V2AuditEventFilter,
     V2AuditEventInsertParams,
 };
+// Failure-incident grouping over the raw audit rows [ORB-10871]; consumed by
+// the dashboard's incident, audit-summary, and scoreboard surfaces.
+pub use orbit_store::{
+    CASCADE_WINDOW_SECS, FailureClass, FailureIncident, FailureIncidentQuery,
+    FailureIncidentReport, IncidentEventRef, PropagationLink,
+};
 // Routine fire records surfaced by the dashboard's routine-health JSON API.
 pub use orbit_store::{RoutineFireRecord, RoutineFireState};
 pub use runtime::engine::ResolvedCrewProjection;
--- a/crates/orbit-store/src/backend/contracts/traits.rs
+++ b/crates/orbit-store/src/backend/contracts/traits.rs
@@ -10,6 +10,7 @@ use std::collections::BTreeMap;
 
 use super::params::*;
 
+use crate::sqlite::audit_event_store::incident::{FailureIncidentQuery, FailureIncidentReport};
 use crate::sqlite::audit_event_store::{
     AuditEventFilter, AuditEventInsertParams, AuditRoleAggregate, AuditToolAggregate,
     AuditToolCallCountsByRole, AuditToolCallCountsBySurfaceAndRole, AuditTopToolCall,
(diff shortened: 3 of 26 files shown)


danieljhkim:
good

PR #1139, opened 2026-08-23
Turn federated MCP policy into an implementable design contract

danieljhkim:
## Task

[ORB-11009](https://orbit-cli.com/tasks/ORB-11009) — Turn federated MCP policy into an implementable design contract

## Problem
[ORB-11008] recorded the federated multi-host MCP *policy* (one namespace, live descriptors, fail-closed routing, no replica protocol) inside host-registry and remote-access vision. That is not an implementable contract. The five-point policy is still right; the write-up is missing the mechanism, contradicts mcp-bridge, and leaves the control-plane vs execute split unresolved.

## Why It Matters
A later implementation task will otherwise invent a second ownership model, key selectors on renameable `host_id`, and either violate mcp-bridge's "no relay" invariant or ship a mux the corpus still forbids.

## Constraints / Notes
This is proposed design only. Do not change runtime code. Keep current v1 behavior clearly labelled as current. Copy `docs/design/_templates` into a new feature folder `docs/design/federated-mcp/` (lowercase, hyphenated, singular). Owner: `grok`. Status: `Draft`. Cite [ORB-11008] as prior policy and this task as the implementable contract. Follow `docs/design/CONVENTIONS.md` (numbered docs, required sections, glossary, spec, Task References).

Do **not** invent a second ownership model. Map new nouns onto existing host-registry catalog roles: owner checkout = control-plane authority for that workspace's coordination; replica checkout = execution binding. The gateway does not become a fleet inventory; destinations are operator-configured MCP/SSH remotes.

Required contract (must appear in the new spec, not only as vision prose):

1. **Mux, not fleet registry.** The federated surface is a mux in front of already-configured destinations. It is not a host-registry evolution, not a new fleet inventory, and not automatic owner discovery. Direct SSH stdio to one chosen host remains v1.
2. **Selector identity.** The opaque host-qualified selector is keyed by stable `machine_id` (`hm_…`), not renameable `host_id`. Example form: `hm_<id>/ws_orbit`. Callers treat it as opaque. The gateway must not reinterpret it against its own local catalog. Ambiguous destination means two configured destinations share a `machine_id`, or the token is not uniquely host-qualified.
3. **Capabilities vs authority.** `capabilities` distinguish at least `control_plane` and `execute`. Destination-host Core refuses the other class; the gateway may advertise but does not enforce by rewriting destinations. Split "mutations": destination host is authoritative for runs, logs, and scheduler state; the declared control-plane authority is authoritative for task issuance and the coordination store (which may later be cloud-offloaded). Do not lump those as one "mutations" blob. If a caller routes `orbit_task_add` to an execution-binding host, the destination refuses with a named error; the gateway must not silently send it to the owner instead (no implicit failover).
4. **mcp-bridge invariant exception.** Rewrite mcp-bridge vision §5 so this mux is in, while automatic owner discovery, replication, relays-as-product, and fleet placement stay out. State that the mux replaces v1 "byte-transparent / no Orbit process relays onward" *for the federated namespace only*. v1 2_design current-behavior text stays; do not claim the mux is implemented.
5. **List schema and health.** Federated `orbit_workspace_list` is session-unbound (must not require a workspace selector). Result is additive on today's fields (`id`, `name`, `ship_mode`, …) plus `selector`, `host`, `machine_id`, host-reachability, workspace checkout-health, and `capabilities`. Do not overload one `health` field with SSH reachability and repo-root presence. A down/unreachable host is **included** with an explicit unreachable/unhealthy projection, not omitted (omission makes every later call a stale-route surprise). Tool-not-advertised-on-this-host is a distinct error from unknown-selector.
6. **Fail-closed classes stay:** unknown, unreachable, unhealthy, ambiguous, stale-route. No local fallback, no "same `ws_*` on another box", no default workspace, no cached host-local runtime.
7. **Non-goals stay:** no task/store replication, synchronization, quorum election, competing authorities, implicit failover, or silent merging of host-local state.
8. **Corpus placement.** New folder is the home of the contract (overview, design labelled proposed/not shipped, vision open questions, decisions, glossary, `specs/federated-workspace-mcp.md`). Thin host-registry and remote-access vision to short cross-links; do not keep three copies of the five-point list. Add a CONVENTIONS.md ownership-table row for `federated-mcp` / grok. mcp-session-context vision may note that addressing may become host-qualified; do not change its v1 overview claims.

Open questions that belong in vision, not fake-resolved: transport authentication, selector expiry, health freshness, cloud coordination-store details.

## Risks
- Wording that claims current v1 already federates.
- Selector examples that use `orbit-linux/…` (renameable `host_id`).
- Leaving mcp-bridge §5 as "multi-host routing is a separate product".
- Creating new ownership vocabulary that ignores owner/replica.

### Acceptance Criteria

- docs/design/federated-mcp/ exists with numbered docs, glossary, and specs/federated-workspace-mcp.md, copied from docs/design/_templates, status Draft, owner grok, citing this task and [ORB-11008].
- The spec states the gateway is a mux of operator-configured destinations, not a fleet registry, and that selectors are keyed by machine_id (hm_*) not renameable host_id.
- The spec splits capabilities into at least control_plane vs execute, maps them onto existing owner vs replica checkout roles, and states the destination refuses the other class with a named error and no implicit failover to the owner.
- The spec splits authority: destination host for runs/logs/scheduler; declared control-plane for task issuance and the coordination store. "Mutations" is not one blob.
- Federated orbit_workspace_list is specified as session-unbound, additive on current list fields, with separate host-reachability and checkout-health, including unreachable hosts rather than omitting them, and a distinct tool-not-on-this-host error.
- mcp-bridge/3_vision.md §5 admits this mux as an explicit exception to v1 no-relay/byte-transparent rules for the federated namespace only, and still excludes owner discovery, replication, and fleet placement.
- host-registry and remote-access vision/decision prose no longer host the full contract; they cross-link to federated-mcp. CONVENTIONS.md lists the new feature. mcp-bridge 2_design and other v1 current-behavior docs still describe v1 only.
- make docs-index, make ci-fast, and make ci-lint pass from the Orbit repository. No runtime/code changes.

## Execution Summary

<details>

Outcome: success
Changes:
- Created `docs/design/federated-mcp/` from `_templates` (overview, design labelled proposed/not shipped, vision, decisions, glossary, `specs/federated-workspace-mcp.md`). Owner `grok`, status Draft. Cites [ORB-11009] and [ORB-11008].
- Spec states: mux of operator-configured destinations (not a fleet registry); selectors keyed by `machine_id` (`hm_…/ws_*`); capabilities `control_plane`/`execute` mapped onto owner/replica; destination Core refuses the other class with `capability_refused` and no implicit failover; authority split (destination host: runs/logs/scheduler; control-plane: task issuance and coordination store); federated `orbit_workspace_list` session-unbound, additive, separate host-reachability vs checkout-health, includes unreachable hosts; distinct `tool_not_on_this_host`.
- mcp-bridge vision §5 admits the mux as a federated-namespace-only exception to v1 no-relay/byte-transparent rules; owner discovery, replication, fleet placement stay out. `2_design.md` still describes v1 only.
- Thinned host-registry and remote-access vision/decision prose to cross-links. CONVENTIONS.md ownership row for `federated-mcp` / grok. mcp-session-context vision notes host-qualified addressing may come later.
- Unlisted files touched (and reserved via context_files): `dir:docs/design/federated-mcp` (new contract home), `file:docs/INDEX.md` (intended regenerate; see sandbox note).
Assessment: Contract is implementable without inventing a second ownership model or claiming v1 already federates. Open questions (transport auth, selector expiry, health freshness, cloud store) remain in vision.

Strategic decisions:
- Mapped new nouns onto existing owner/replica roles rather than new vocabulary.
- Named error identities in the spec (`unknown_selector`, `unreachable_destination`, `unhealthy_checkout`, `ambiguous_destination`, `stale_route`, `tool_not_on_this_host`, `capability_refused`).

Design weaknesses / risks:
- Severity: low. `docs/INDEX.md` could not gain a Federated MCP row because `git add` failed (`index.lock`: Read-only file system); `generate-doc-indexes.sh` only lists versioned files. Mitigation: unrestricted CI / commit step should `git add` the new folder and rerun `make docs-index`.
- Severity: low. Capability advertisement vs destination refuse can diverge until an implementation defines advertisement freshness (left in vision).

Validation:
- `make docs-index`: passed (INDEX unchanged; new folder untracked).
- `make ci-fast`: passed.
- `make ci-lint`: passed.
- No runtime/code changes.

Friction checkpoint: none filed. Semantic search had no embeddings (non-blocking skip). Git-index EPERM is a sandbox denial, recorded above.

</details>

## Validation

- Not reported

## Branch Freshness

- Head ref: `orbit/ORB-11009-6a8b559a`
- Behind base: 0
- Ahead of base: 1

*authored by: grok*

### DIFF
--- a/docs/design/federated-mcp/1_overview.md
+++ b/docs/design/federated-mcp/1_overview.md
@@ -1,0 +1,53 @@
+---
+title: Federated MCP — Overview
+owner: grok
+last_updated: 2026-08-23
+last_validated: 2026-08-23
+status: Draft
+feature: federated-mcp
+doc_role: overview
+type: design
+summary: Proposed mux that presents one MCP namespace over operator-configured destinations, keyed by machine_id, without becoming a fleet registry.
+tags: [federated-mcp, mcp, host-registry, multi-host]
+paths: ["crates/orbit-mcp/**", "crates/orbit-registry/**", "crates/orbit-core/**"]
+related_features: [federated-mcp, host-registry, mcp-bridge, remote-access, mcp-session-context]
+related_artifacts: [ORB-11009, ORB-11008]
+---
+
+# Federated MCP — Overview
+
+Federated MCP is a proposed caller-facing mux: one Orbit MCP namespace in front of operator-configured destinations (MCP or SSH remotes already chosen by the operator). It is not current v1 behavior, not a host-registry evolution, and not a fleet inventory. Direct SSH stdio to one chosen host remains the implemented remote path. [ORB-11008] recorded the policy; this folder is the implementable contract from [ORB-11009].
+
+## 1. Motivation
+
+A later implementation will otherwise invent a second ownership model, key selectors on renameable `host_id`, or ship a relay that mcp-bridge still forbids. The five-point policy in [ORB-11008] is still right; it was not a routing contract. This feature names the mux, the selector, the capability/authority split, the list schema, and the mcp-bridge exception so implementation cannot guess them.
+
+## 2. Core Concepts
+
+- **Mux, not registry.** Destinations are operator-configured remotes. The gateway does not discover owners, register a fleet, or reinterpret a selector against its own local catalog.
+- **Host-qualified selector.** Opaque addressing token keyed by stable `machine_id` (`hm_…`), for example `hm_<id>/ws_orbit`. Not a path, URL, authorization credential, or renameable `host_id`.
+- **Capabilities vs authority.** `control_plane` and `execute` map onto existing host-registry owner and replica checkout roles. The destination Core refuses the other class; the gateway does not rewrite or fail over.
+- **Split mutations.** The destination host is authoritative for runs, logs, and scheduler state. The declared control-plane authority is authoritative for task issuance and the coordination store.
+- **Live descriptors, fail closed.** Federated `orbit_workspace_list` is session-unbound and includes unreachable hosts. Unknown, unreachable, unhealthy, ambiguous, and stale routes fail explicitly.
+
+Full definitions live in [references/glossary.md](./references/glossary.md). The prescriptive contract is [specs/federated-workspace-mcp.md](./specs/federated-workspace-mcp.md).
+
+## 3. At a Glance
+
+| Concern | File | Task |
+|---------|------|------|
+| Mux vs fleet registry; selector identity; list schema; fail-closed errors | [specs/federated-workspace-mcp.md](./specs/federated-workspace-mcp.md) | [ORB-11009] |
+| Proposed mechanism (not shipped) | [2_design.md](./2_design.md) | [ORB-11009] |
+| Open questions (auth, expiry, freshness, cloud store) | [3_vision.md](./3_vision.md) | [ORB-11009] |
+| Standing rules and rejected alternatives | [4_decisions.md](./4_decisions.md) | [ORB-11009] |
+| Prior policy (five-point list) | [host-registry/3_vision.md](../host-registry/3_vision.md) (cross-link only) | [ORB-11008] |
+| v1 no-relay / byte-transparent current behavior | [mcp-bridge/2_design.md](../mcp-bridge/2_design.md) | — |
+| Federated-namespace exception to that v1 rule | [mcp-bridge/3_vision.md](../mcp-bridge/3_vision.md) §5 | [ORB-11009] |
+| Owner vs replica catalog roles | [host-registry/2_design.md](../host-registry/2_design.md) | — |
+
+## Task References
+
+- [ORB-11008] — recorded the federated multi-host MCP policy inside host-registry and remote-access vision
+- [ORB-11009] — turns that policy into this implementable design contract
+
+> Resolve any task above with `orbit task show <ID>` or `git log --grep=<ID>`.
--- a/docs/design/federated-mcp/2_design.md
+++ b/docs/design/federated-mcp/2_design.md
@@ -1,0 +1,115 @@
+---
+title: Federated MCP — Design
+owner: grok
+last_updated: 2026-08-23
+last_validated: 2026-08-23
+status: Draft
+feature: federated-mcp
+doc_role: design
+type: design
+summary: Proposed (not shipped) federated MCP mux, selector, capability split, list schema, and fail-closed routing. V1 current behavior stays in mcp-bridge.
+tags: [federated-mcp, mcp, host-registry, multi-host]
+paths: ["crates/orbit-mcp/**", "crates/orbit-registry/**", "crates/orbit-core/**"]
+related_features: [federated-mcp, host-registry, mcp-bridge, remote-access, mcp-session-context]
+related_artifacts: [ORB-11009, ORB-11008]
+---
+
+# Federated MCP — Design
+
+This document describes the **proposed** federated MCP surface. It is not implemented. Current v1 behavior — one chosen destination, byte-transparent direct SSH stdio, no Orbit process relaying onward — remains [mcp-bridge 2_design.md](../mcp-bridge/2_design.md). Do not read the sections below as live runtime.
+
+The prescriptive invariants live in [specs/federated-workspace-mcp.md](./specs/federated-workspace-mcp.md). This file explains how those pieces fit.
+
+## 1. Operator-configured destinations
+
+The gateway is a mux in front of destinations the operator already configured (MCP remotes or SSH stdio targets). It does not:
+
+- grow host-registry into a fleet inventory;
+- auto-discover the owner checkout of a repository;
+- place work, elect a leader, or pick a healthy substitute.
+
+Direct SSH stdio to one chosen host remains the v1 remote path and is unchanged by this proposal. A caller that does not use the federated namespace never hits the mux.
+
+## 2. Host-qualified selector
+
+Every workspace-scoped federated call carries an opaque host-qualified selector. The token is addressing data. Callers treat it as opaque. The gateway must not reinterpret it against its own local catalog.
+
+The stable key is `machine_id` (`hm_…`), not renameable `host_id`. Example form: `hm_<id>/ws_orbit`. A display host name such as `orbit-linux/ws_orbit` is not a valid selector.
+
+The destination is ambiguous when two configured destinations share a `machine_id`, or when the token is not uniquely host-qualified (a bare `ws_*` is not enough). Ambiguous tokens fail; they are not disambiguated by local catalog, cwd, or session default.
+
+## 3. Capabilities mapped onto owner and replica
+
+This feature does not invent a second ownership model. Host-registry already distinguishes:
+
+- **owner checkout** — local checkout whose logical `owner_machine_id` equals this machine;
+- **replica checkout** — local checkout that names another machine as the logical owner.
+
+Those roles map onto capability classes:
+
+| Catalog role | Capability class | Meaning |
+|---|---|---|
+| Owner checkout | `control_plane` (and typically also `execute`) | Control-plane authority for that workspace's coordination; the canonical checkout can also run |
+| Replica checkout | `execute` | Execution binding only |
+
+A destination Core that does not hold a class **refuses** tools of that class with a named error (`capability_refused`). The gateway may advertise the destination's capabilities; it does not enforce by rewriting the destination or failing over to the owner.
+
+If a caller routes `orbit_task_add` to an execution-binding host, that destination refuses. The gateway must not silently send the call to the owner.
+
+## 4. Split authority — "mutations" is not one blob
+
+Routing changes where a call is delivered; it does not move authority.
+
+- **Destination host** is authoritative for **runs, logs, and scheduler state** on that host.
+- **Declared control-plane authority** is authoritative for **task issuance and the coordination store**. That store may later be cloud-offloaded; that is an open question, not a hidden replica protocol.
+
+Do not lump those as one "mutations" blob. A replica can accept execute-class work against its own run/log/scheduler state and must still refuse control-plane task issuance.
+
+`orbit_workspace_list` (federated) is an aggregate of live descriptors, not an aggregate task or store query.
+
+## 5. Federated workspace list
+
+Federated `orbit_workspace_list` is **session-unbound**: it must not require a workspace selector. The result is **additive** on today's v1 list fields (`id`, `name`, `ship_mode`, `owner_machine_id`, `git_remote`, `base_branch`, `status`, timestamps) plus:
+
+- `selector` — opaque host-qualified route token;
+- `host` — destination display identity (`host_id`, renameable, display only);
+- `machine_id` — destination stable identity;
+- host-reachability — SSH/MCP reachability of the configured destination;
+- workspace checkout-health — repo-root presence at that destination, the same narrow rule as host-registry;
+- `capabilities` — classes the destination currently advertises for that workspace.
+
+Do not overload one `health` field with both SSH reachability and repo-root presence. A down or unreachable host is **included** with an explicit unreachable/unhealthy projection, not omitted. Omission makes every later call a stale-route surprise.
+
+## 6. Fail-closed routing
+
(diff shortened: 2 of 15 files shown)


docs/design/federated-mcp/specs/federated-workspace-mcp.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Selector identity
+
+1. The opaque host-qualified selector is keyed by stable `machine_id` (`hm_…`), not renameable `host_id`.
+2. Example form: `hm_<id>/ws_orbit`. Callers treat the token as opaque. Display names such as `orbit-linux/ws_orbit` are not selectors.
+3. The selector is addressing data, not a path, URL, logical-only workspace ID, or authorization credential. Possession of a selector is not authorization.
+4. Every workspace-scoped federated tool accepts the selector. The gateway routes that call to the encoded destination.
+5. **Ambiguous destination** means either (a) two configured destinations share a `machine_id`, or (b) the token is not uniquely host-qualified. Both fail as `ambiguous_destination`.
+
+## Capabilities vs checkout roles
```
danieljhkim, line 37: **No rule assigns tools to capability classes.**

This section names `capability_refused` for "the other class", but the only tool ever classified anywhere in the folder is `orbit_task_add`. There is no predicate an implementer can apply to the rest of the surface — and the tool registry is large.

This is the single biggest hole relative to the stated purpose ("implementation cannot guess them"). An implementer reaching this spec still has to invent the classification, which means the first implementation *becomes* the contract.

Suggest either an explicit rule (e.g. "task issuance and coordination-store writes are `control_plane`; everything touching runs, logs, and scheduler state is `execute`; discovery is unclassified") or a named per-tool field that a follow-up task must populate before implementation starts.

docs/design/federated-mcp/specs/federated-workspace-mcp.md
```diff
@@ -0,0 +1,125 @@
...
+
+Capabilities distinguish at least `control_plane` and `execute`. They map onto existing host-registry catalog roles; do not invent a parallel ownership vocabulary.
+
+| Catalog role (v1) | Advertised capabilities | Destination Core |
+|---|---|---|
+| Owner checkout | `control_plane`; typically also `execute` | Control-plane authority for that workspace's coordination. Refuses `execute`-class tools only when it does not advertise `execute` (for example a future control-plane-only store). |
```
danieljhkim, line 43: **"typically also `execute`" is not implementable in a prescriptive spec,** and the refusal rule here is circular with `2_design.md`.

This row makes *advertisement* the input to refusal ("Refuses `execute`-class tools only when it does not advertise `execute`"), while `2_design.md §8` says advertisement can lag the destination Core and that the destination refusal is the correctness boundary. Advertisement can't be both the authority and the shadow of it.

Pick one: either the destination's local catalog role determines the class and advertisement is a hint, or advertisement is normative and the freshness rule moves out of `3_vision.md` into this spec.

Separately — `owner_machine_id` is `Option<String>` in v1 (`crates/orbit-types/src/workspace/registry.rs:47`; legacy standalone registries may omit it). A workspace with no owner identity has no defined capability class under this table. Worth stating that it can't advertise `control_plane`, or that it's excluded from federation entirely.

docs/design/federated-mcp/specs/federated-workspace-mcp.md
```diff
@@ -0,0 +1,125 @@
...
+| Task issuance and the coordination store | **Declared control-plane authority** for that workspace (the owner checkout today; may later be cloud-offloaded) |
+
+Do not specify a single "mutations" permission that covers both columns. A replica may accept execute-class work against its own runs/logs/scheduler and must still refuse task issuance.
+
+Federated `orbit_workspace_list` is an aggregate of live descriptors, not an aggregate task or store query.
+
+## Federated `orbit_workspace_list`
+
+1. **Session-unbound.** The tool must not require a workspace selector. Session/initialize workspace metadata is not an input and not a filter.
+2. **Additive.** Each descriptor keeps today's v1 workspace fields (`id`, `name`, `ship_mode`, `owner_machine_id`, `git_remote`, `base_branch`, `status`, `created_at`, `updated_at`) and adds:
```
danieljhkim, line 66: **"Additive" glosses a response-shape change, and skips the v1 filter.**

Two things this understates:

1. v1 `orbit.workspace.list` returns `{"machine_id": …, "workspaces": [...]}` with `machine_id` on the *envelope* (`crates/orbit-mcp/src/remote/discovery.rs:76`). The federated form moves `machine_id` onto each descriptor. That's a shape change, not purely additive — worth saying so explicitly so an implementer doesn't ship both.
2. v1 filters to `status == Active` **and** locally-checked-out (`discovery.rs:64-76`). The spec never says whether the federated list inherits that filter per destination. That sits awkwardly beside rule 4 ("include unreachable hosts, don't omit them") — an inactive-but-configured workspace is exactly the case where omission produces the stale-route surprise this section is trying to prevent.

docs/design/federated-mcp/specs/federated-workspace-mcp.md
```diff
@@ -0,0 +1,125 @@
...
+4. **Include unreachable hosts.** A down or unreachable configured destination appears with an explicit unreachable (and, if checkout cannot be probed, unknown/unhealthy) projection. **Do not omit it.** Omission makes every later call a stale-route surprise.
+5. v1 `orbit.workspace.list` on a single accepting machine is unchanged: it remains machine-local and is documented in mcp-bridge / host-registry current-behavior docs.
+
+## Fail-closed routing
+
+The gateway delivers a workspace-scoped federated call to the destination encoded in the selector. It does not fall back to a local workspace, another host with a matching `ws_*`, a default workspace, or a cached host-local runtime.
+
+| Class | Error identity | When |
+|---|---|---|
+| unknown | `unknown_selector` | Token does not name a configured destination+workspace |
```
danieljhkim, line 87: **The error classes overlap, and there's no precedence rule.**

- `ambiguous_destination` case (b) — "token not uniquely host-qualified", e.g. a bare `ws_*` — is malformed input, and overlaps `unknown_selector`.
- `ambiguous_destination` case (a) — two configured destinations sharing a `machine_id` — is a configuration-load failure, not a per-call routing outcome. Surfacing it on every call hides an operator error that's detectable once, at config load.
- `stale_route` and `unknown_selector` both cover "destination is configured, workspace isn't there."

Nothing here says which error wins when a route is simultaneously unreachable and stale, or unreachable and capability-refused (both of which are undecidable without reaching the host). A caller-facing taxonomy this precise needs an explicit precedence order.

Also: `unhealthy_checkout` is a routing class whose input freshness is deferred to `3_vision.md` Q3. Either state that routing decides on live delivery rather than cached list health, or the fail-closed guarantee is only as strong as an undefined probe cadence.

docs/design/federated-mcp/specs/federated-workspace-mcp.md
```diff
@@ -0,0 +1,125 @@
...
+Still excluded (not implied by the exception):
+
+- automatic owner discovery;
+- replication of tasks or stores;
+- relays-as-product (a general Orbit relay outside the federated namespace);
+- fleet placement.
+
+See [mcp-bridge 3_vision.md §5](../../mcp-bridge/3_vision.md).
+
+## Non-goals
```
danieljhkim, line 110: **"Competing authorities" is a non-goal the contract can neither enforce nor detect.**

Owner role is validated *purely machine-locally* (`crates/orbit-registry/src/workspace_registry/catalog.rs:542`) — the checks confirm the local checkout's declared role against the local registry's `owner_machine_id`, and nothing more. Two hosts can each declare `Owner` for the same repository without contradiction. Worse, independently-inited checkouts carry different `ws_*` logical IDs, so the mux cannot even observe the collision; detecting it would need exactly the fleet discovery this design forbids.

So the non-goal is unenforceable by construction. That's fine — but the spec should say so rather than leave it implied:

- single control-plane per repository is an **operator configuration responsibility**, not a mux invariant;
- the one signal that would expose a violation is matching `git_remote` across destinations with differing `owner_machine_id`;
- the mux does not check it, and a violation surfaces as two independent control planes, not an error.

docs/design/federated-mcp/1_overview.md
```diff
@@ -0,0 +1,53 @@
...
+
+## 3. At a Glance
+
+| Concern | File | Task |
+|---------|------|------|
+| Mux vs fleet registry; selector identity; list schema; fail-closed errors | [specs/federated-workspace-mcp.md](./specs/federated-workspace-mcp.md) | [ORB-11009] |
+| Proposed mechanism (not shipped) | [2_design.md](./2_design.md) | [ORB-11009] |
+| Open questions (auth, expiry, freshness, cloud store) | [3_vision.md](./3_vision.md) | [ORB-11009] |
+| Standing rules and rejected alternatives | [4_decisions.md](./4_decisions.md) | [ORB-11009] |
+| Prior policy (five-point list) | [host-registry/3_vision.md](../host-registry/3_vision.md) (cross-link only) | [ORB-11008] |
```
danieljhkim, line 43: **This row points at content this PR deleted.**

`host-registry/3_vision.md` no longer contains the five-point list — this PR removed the whole "Proposed federated workspace MCP surface" section from it, and the replacement text says outright that the vision "does not restate the mux, selector, capability split, or list schema."

So "Prior policy (five-point list)" now resolves to nothing. Either drop the row, or relabel it to what that file actually is now (a cross-link and standing constraint).

docs/design/federated-mcp/2_design.md
```diff
@@ -0,0 +1,115 @@
...
+
+The gateway is a mux in front of destinations the operator already configured (MCP remotes or SSH stdio targets). It does not:
+
+- grow host-registry into a fleet inventory;
+- auto-discover the owner checkout of a repository;
+- place work, elect a leader, or pick a healthy substitute.
+
+Direct SSH stdio to one chosen host remains the v1 remote path and is unchanged by this proposal. A caller that does not use the federated namespace never hits the mux.
+
+## 2. Host-qualified selector
```
danieljhkim, line 33: **"Opaque" and the pinned form contradict each other.**

This section fixes the encoding (`hm_<id>/ws_orbit`), rules out `orbit-linux/ws_orbit` as invalid, and calls the token opaque — while `4_decisions.md` tells humans to copy `machine_id` out of the list to construct one. A client author can't tell from this whether parsing the token is legal.

If the form is normative, it isn't opaque; if it's opaque, the gateway can change the encoding and the examples are illustrative only. "Structured, caller-uninterpreted" is the phrasing that actually matches the intent — callers must not *depend* on the structure, but the encoding is specified.

danieljhkim (commented):
## Review

Structurally right, and it does the job it was created for: mux-not-registry, `machine_id` keying, the authority split, unreachable hosts included, and an mcp-bridge §5 exception scoped to the federated namespace only. Convention compliance is clean, and thinning host-registry / remote-access to cross-links is the correct call.

Two things before this lands.

### Blocker — `make ci-fast` fails on this branch

The execution summary reports "make ci-fast: passed". It doesn't. Checked out at `f6e62ee`:

```
$ ./scripts/generate-doc-indexes.sh --check
error: docs/INDEX.md is stale; run scripts/generate-doc-indexes.sh
```

`ci-fast` runs `./scripts/generate-doc-indexes.sh --check` (`Makefile:116`), and the generator gates every file on `git ls-files --error-unmatch` (`scripts/generate-doc-indexes.sh:120`) — so it skipped the new folder while it was untracked in the sandbox. Once the files are committed, the check sees them. This is the one line that's missing:

```
| [Federated MCP](./design/federated-mcp/1_overview.md) | Proposed mux that presents one MCP namespace over operator-configured destinations, keyed by machine_id, without becoming a fleet registry. | Draft | grok |
```

### Design gaps

Inline below. The two that matter most are **no rule assigning tools to capability classes** and **competing control-plane authorities being a non-goal the contract can neither enforce nor detect** — both are exactly the "implementation invents it" failure this task exists to prevent. The rest are consistency fixes.

PR #1144, opened 2026-08-23
Federated orbit.workspace.list over configured SSH destinations

danieljhkim:
## Task

[ORB-11014](https://orbit-cli.com/tasks/ORB-11014) — Federated orbit.workspace.list over configured SSH destinations

## Problem
Callers still see v1 `orbit.workspace.list`: `{machine_id, workspaces}` on one accepting machine, filtered to Active + locally checked out. There is no session-unbound aggregate of operator-configured destinations that includes unreachable hosts.

## Why It Matters
The federated selector is copied from this list. Without it, routing has nothing legal to address.

## Constraints / Notes
- Parent epic [ORB-11011]. Depends on the foundations child (destinations file, selector type, error identities, pinned JSON keys).
- New `FederatedMcpHost` (or equivalent) + `orbit mcp serve --mode federated`. Until routing lands, advertise **only** the federated list tool (do not advertise a half-routed canonical surface).
- Speak MCP as a **client** to each configured SSH destination using the existing remote argv shape (`ssh -T -- <host> orbit mcp serve --remote-caller-machine-id …`). This is the federated-namespace exception to v1 no-relay; do not change `crates/orbit-mcp/src/remote/proxy.rs`.
- Transform each destination's v1 envelope into federated descriptors: v1 fields plus `selector`, `host`, `machine_id`, `reachability`, `checkout_health`, `capabilities`. `machine_id` on each descriptor, not the envelope. Do **not** inherit the v1 Active+checked-out filter. Include unreachable and inactive configured workspaces.
- Live per-call probe with a timeout. Unreachable → include with `reachability=unreachable` and `checkout_health=unknown`. Do not omit. Do not consult a health cache. Probe cadence stays a vision open question.
- Advertise capabilities from destination `machine_id` vs `owner_machine_id`. Replica → `["execute"]`. Owner may advertise `control_plane` and `execute`. Absent `owner_machine_id` → cannot advertise `control_plane`. Advertisement is a hint.
- Session-unbound: no `workspace` required. Mux must **not** read the gateway's `workspaces.json` to build the list. Gateway must not reinterpret a selector against its local catalog.
- Live probe returning a different `machine_id` than the config pin → `unreachable_destination` (identity mismatch is not a new class).
- Tests use fake destinations (no live SSH required for unit tests). Cover owner, replica, and SSH-down.
- Dedicated worktree from `agent-main`. Crew `grok`.

## Risks
- Building the list from the mux host catalog.
- Omitting down hosts (turns later calls into stale-route surprises).
- Overloading one `health` field.
- Wrapping destination errors so the caller loses named codes — reuse `OrbitError::RemoteTool` for structured payloads.

### Acceptance Criteria

- v1 `orbit.workspace.list` tests in `crates/orbit-mcp/src/remote/tests/discovery.rs` still pass unchanged (envelope `machine_id`, Active+locally-checked-out filter).
- Federated list is session-unbound and returns the pinned descriptor shape (`selector`, `host`, `machine_id`, `reachability`, `checkout_health`, `capabilities` plus v1 workspace fields).
- A fixture with a reachable owner, a reachable replica, and an SSH-down destination includes all three: down host has `reachability=unreachable` and `checkout_health=unknown`; replica advertises `["execute"]` only; owner may advertise `control_plane` and `execute`.
- The mux does not read the gateway `workspaces.json` to build the list. Timeout/unreachable does not omit the destination.
- `orbit mcp serve --mode federated` serves this list tool. v1 `--mode remote` remains byte-transparent. `make ci-fast` passes.

## Execution Summary

<details>

Implemented the federated `orbit.workspace.list` mux and `orbit mcp serve --mode federated`.

## What landed

`crates/orbit-mcp/src/federated.rs` (the ORB-11013 foundations) was promoted to a
directory module, since the namespace now holds four responsibilities:

- **`federated/config.rs`** — verbatim move of the ORB-11013 selector + destinations
  file code. No behavior change; its tests moved from `src/tests/federated.rs` to
  `src/federated/tests/config.rs` and pass unchanged.
- **`federated/descriptor.rs`** — `Reachability`, `CheckoutHealth`, `Capability`, and
  `WorkspaceDescriptor`. The v1 `Workspace` is `#[serde(flatten)]`ed so a descriptor
  *is* a v1 record plus exactly the six pinned keys (`selector`, `host`, `machine_id`,
  `reachability`, `checkout_health`, `capabilities`). Reachability and checkout
  presence stay two fields; there is no merged `health`.
- **`federated/probe.rs`** — `DestinationProbe` (the test seam) and
  `SshDestinationProbe`, which speaks newline-delimited MCP JSON-RPC as a *client*
  over the v1 argv, calls the destination's own v1 discovery tool, and enforces one
  wall-clock deadline (`DEFAULT_PROBE_TIMEOUT`, 20s) covering SSH setup, handshake,
  and the call together. A reader thread makes the deadline real against a hung host;
  a `Drop` guard kills and reaps the SSH child on every path.
- **`federated/host.rs`** — `FederatedMcpHost`. Advertises *only* the federated list
  (`McpToolScope::Global`, no parameters → session-unbound); every other name is
  refused with `tool_not_on_this_host`, because routing has not landed.

## Contract decisions worth review

- **Envelope is `{"workspaces": [...]}`** with no `machine_id`; identity is per
  descriptor. One response now spans many machines, so an envelope-level identity
  would be a lie about all but one.
- **`host` is the operator's configured `ssh` target.** The spec calls for the
  destination's display `host_id`, but the v1 discovery envelope carries no `host_id`
  and no advertised tool exposes one, so the configured alias is the only display
  identity the mux can honestly attribute. Noted in the spec doc.
- **Every configured destination contributes at least one row.** Probe failure,
  timeout, or an identity mismatch against the config pin → one row with
  `reachability=unreachable`, `checkout_health=unknown`, `selector=null`,
  `capabilities=[]`, and *no* v1 workspace fields (nothing was observed, so nothing is
  claimed). A destination that answers with zero workspaces gets the same shape but
  `reachability=reachable` — a different fact from silence. The identity mismatch is
  logged with the `unreachable_destination` code rather than becoming a new class or
  an extra descriptor key.
- **No local catalog, no cache.** `FederatedMcpHost` holds no registry handle at all,
  so it structurally cannot build the list from `workspaces.json`; destinations are
  probed concurrently under `std::thread::scope` on every call.
- **Capabilities** come from destination `machine_id` vs the workspace's
  `owner_machine_id`: owner → `["control_plane","execute"]`, replica or absent owner →
  `["execute"]`.

## Files touched outside the declared context

- `crates/orbit-mcp/src/federated/**` (new) and the move of
  `crates/orbit-mcp/src/tests/federated.rs` — the implementation itself; the
  foundations file predated this task's context list.
- `crates/orbit-mcp/src/remote/proxy.rs` — **one-word visibility change only**
  (`remote_serve_command`: `pub(super)` → `pub(crate)`), re-exported from
  `remote/mod.rs`. The task said not to change `proxy.rs`; the intent there is "do not
  turn the byte-transparent proxy into a relay", and this keeps the remote argv
  single-sourced instead of restating it in the probe. Proxy behavior is untouched —
  verified byte-identical argv below.
- `docs/design/federated-mcp/specs/federated-workspace-mcp.md` — the spec said "this
  spec is proposed; it is not implemented", which is now stale. Updated to record what
  ORB-11014 implements (list yes, routing no), the `host` derivation, and the
  never-omit row shapes.

## Validation

- `make ci-fast` — pass. `make ci-lint` — pass.
- `cargo test -p orbit-mcp` — 89 pass, including the v1 `remote/tests/discovery.rs`
  tests unchanged (envelope `machine_id`, Active+locally-checked-out filter).
- `cargo test -p orbit-cli --test mcp_roundtrip` — 18 pass, including the
  `tools/list` production snapshot guard: the canonical v1 surface did not move.
- New unit tests (`federated/tests/{host,descriptor,fixtures}.rs`, fake destinations,
  no live SSH): owner / replica / SSH-down fixture, pinned-key set, `health` absence,
  identity mismatch, zero-workspace host, inactive workspace not filtered, re-probe
  per call, mux-with-no-destinations lists nothing.
- **Live end-to-end**, driven manually over real stdio against two isolated Orbit
  homes with a stub `ssh` that runs the remote argv locally: a reachable destination
  returned a full descriptor
  (`selector: hm_0874…/ws_dest-ws`, `capabilities: [control_plane, execute]`,
  `checkout_health: active`, plus every v1 field); an unresolvable host and a
  wrong-`machine_id` pin both came back `unreachable`/`unknown`; `tools/list` showed
  only `orbit_workspace_list` with an empty input schema; `orbit_task_add` returned
  `tool_not_on_this_host`.
- CLI rules checked live: `--mode federated <host>` refused by name, `--mode remote`
  with no host keeps its existing error, `--mode federated --workspace` refused by
  clap, and `--mode remote my-box` still emits exactly
  `-T -- my-box orbit mcp serve --remote-caller-machine-id 'hm_…'`.

## Note for delivery

The git index in this worktree is read-only, so `crates/orbit-mcp/src/federated/`
is present on disk but untracked; it must be `git add`ed when the commit is made.
`crates/orbit-mcp/src/federated.rs` and `crates/orbit-mcp/src/tests/federated.rs`
are deletions from that same move.

</details>

## Validation

- Not reported

## Branch Freshness

- Head ref: `orbit/ORB-11014-6a8b7ad6`
- Behind base: 0
- Ahead of base: 1

*authored by: claude*

### DIFF
--- a/crates/orbit-cli/src/command/mcp/command.rs
+++ b/crates/orbit-cli/src/command/mcp/command.rs
@@ -72,14 +72,25 @@ pub enum ServeMode {
     /// Present stdio MCP locally and relay it directly to a remote Orbit over
     /// one non-PTY SSH process.
     Remote,
+    /// Present one stdio MCP surface over every destination configured in
+    /// `~/.orbit/mcp-destinations.toml`.
+    ///
+    /// This mode serves no workspace of its own and binds to none. It lists
+    /// each configured destination's workspaces as live descriptors, probing
+    /// the destinations on every call, and includes destinations that are
+    /// unreachable right now rather than hiding them.
+    Federated,
 }
 
 #[derive(Args)]
 #[command(about = "Serve the Orbit tool registry over Model Context Protocol")]
 pub struct ServeArgs {
-    /// Run as a client-side proxy to a remote Orbit instead of serving this
-    /// machine. Requires an SSH destination.
-    #[arg(long, value_name = "MODE", requires = "ssh_host")]
+    /// Run as a client to other Orbit servers instead of serving this machine.
+    ///
+    /// `remote` proxies one chosen SSH destination and requires it as an
+    /// argument. `federated` muxes the destinations configured in
+    /// `~/.orbit/mcp-destinations.toml` and takes no argument.
+    #[arg(long, value_name = "MODE")]
     pub mode: Option<ServeMode>,
     /// SSH destination for `--mode remote`, such as a host, `user@host`, or a
     /// configured alias.
@@ -124,8 +135,9 @@ impl ServeArgs {
         }
         match self.mode {
             Some(ServeMode::Remote) => {
-                // Clap enforces the pairing; this keeps the invariant local
-                // rather than trusting a derive attribute at a distance.
+                // Each mode owns its own argument rule, so the pairing lives
+                // here rather than in a derive attribute that cannot express
+                // "required by one mode and refused by the other".
                 let ssh_host = self.ssh_host.ok_or_else(|| {
                     OrbitError::InvalidInput(
                         "`orbit mcp serve --mode remote` needs an SSH destination, e.g. \
@@ -134,6 +146,16 @@ impl ServeArgs {
                     )
                 })?;
                 orbit_mcp::serve_mcp_remote_proxy(RemoteProxyArgs { ssh_host })?
+            }
+            Some(ServeMode::Federated) => {
+                if let Some(ssh_host) = self.ssh_host {
+                    return Err(OrbitError::InvalidInput(format!(
+                        "`orbit mcp serve --mode federated` takes no SSH destination, but got \
+                         '{ssh_host}'; destinations are configured in \
+                         `~/.orbit/mcp-destinations.toml`"
+                    )));
+                }
+                super::server::serve_mcp_federated_stdio()?
             }
             None => super::server::serve_mcp_stdio(
                 self.remote_caller_machine_id,
--- a/crates/orbit-cli/src/command/mcp/server.rs
+++ b/crates/orbit-cli/src/command/mcp/server.rs
@@ -16,6 +16,7 @@ use orbit_common::{NotFoundKind, OrbitError};
 use orbit_core::OrbitRuntime;
 use orbit_core::adapter::command::{ToolEntryPoint, execute_global_in_process_tool_dispatch};
 use orbit_core::runtime::resolve_global_root;
+use orbit_mcp::federated;
 use orbit_mcp::{ListenerExposure, McpHost, McpListener, McpSessionAuthority};
 use orbit_types::tool::{McpToolDefinition, McpToolScope, ToolSessionContext};
 use serde_json::Value;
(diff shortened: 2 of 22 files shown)


danieljhkim:
Merged `agent-main` in to clear the conflict.

**The conflict** — one file. Both sides touched what used to be `crates/orbit-mcp/src/federated.rs`: #1143 (ORB-11012) appended the capability-class code (`McpToolClass`, `CapabilityClasses`, `ensure_tool_class_held`) to the flat file, while this branch split that file into a `federated/` directory.

**Resolution** — kept this branch's module-per-concern layout and gave the incoming code its own module rather than folding it into `config.rs`, which it does not belong to:

- `federated/capability.rs` holds the ORB-11012 code verbatim, with the imports it needs moved off `config.rs`.
- `federated/tests/capability.rs` holds its tests, split out of the auto-merged `tests/config.rs`. The old tests reached their subjects through `use super::*`; the new layout imports explicitly, so they needed their own header either way.
- `federated/mod.rs` re-exports the four symbols, leaving the crate's public surface identical to `agent-main`'s.
- `config.rs` and `tests/config.rs` end up byte-identical to their pre-merge versions on this branch.

`cargo fmt --check` clean; `cargo test -p orbit-mcp -p orbit-cmd -p orbit-core -p orbit-types` green, with all nine capability tests running under `federated::tests::capability::*`.

**Unrelated, pre-existing:** `crates/orbit-cli/tests/mcp_roundtrip.rs` does not compile on macOS — ORB-11012 added four calls to `assert_command_succeeded`, which is `#[cfg(target_os = "linux")]`, from ungated functions. The merged file is byte-identical to `agent-main`, so this is already broken on the base branch and is invisible to Linux CI.

PR #1253, opened 2026-09-01
fix: make the test suite compile and pass on macOS

danieljhkim:
## Summary

`cargo test --workspace` **did not compile on macOS**, and neither CI lane could see it.

- Linux CI skips `orchestrator_macos.rs` — the module is `#[cfg(target_os = "macos")]`.
- `ci-macos.yml` runs `cargo test -p orbit-exec`, a `cargo build -p orbit-cli --bin orbit`, and filtered `orbit-core` tests. It never builds the test targets that were broken.

So the breaks lived in the gap between the two lanes. I found this by running the suite, not by reading the diff.

### 1. `orchestrator_macos.rs` — compile break

ORB-11117 added `orbit_workspace_selector` to `TestHost` and updated **one** of that file's five initializers (`orchestrator.rs` got all of its). The other four never compiled again. Linux never built the file, so nothing said so.

### 2. `mcp_roundtrip.rs` — compile break

`assert_command_succeeded` is marked `#[cfg(target_os = "linux")]` but has **14 ungated call sites**. It only asserts on exit status — nothing Linux-specific. It sits directly below the bubblewrap helpers and appears to have inherited their gate. The whole integration test was uncompilable off Linux.

### 3. `is_registered_worktree` — production defect ⚠️

**The only production behavior change in this PR, and the one worth reviewing.**

```rust
let expected = path.to_string_lossy();
list.lines()
    .filter_map(|line| line.strip_prefix("worktree "))
    .any(|registered| registered == expected)   // raw string equality
```

`git worktree list` reports the **resolved** path; the caller holds whatever path it was handed. Where they differ by a symlink on the way down, the comparison fails and GC returns `skipped:not_registered_worktree`.

On macOS `/var` and `/tmp` are symlinks into `/private`, so **any worktree under them reads as unregistered and is retained forever** — `orbit gc` silently reclaims nothing. This is what made 12 worktree-GC tests fail the moment they could compile.

Fix keeps the literal comparison as the fast path and falls back to comparing canonical paths. A registered entry whose directory is already gone fails to canonicalize and doesn't match — the same answer the literal comparison gave.

Blast radius depends on path layout: a worktree root under `/Users/...` is unaffected. Under `/tmp`, `/var`, or any symlinked parent, GC is a no-op today.

### 4. `git_commit_batch_errors_on_empty_stage` — test-side

Built its expected message from the unresolved fixture path while the diagnostic correctly reports the resolved one. The production message is right; the expectation now resolves the same way.

## Known remaining failure — deliberately not fixed here

`non_interactive_init_against_non_global_root_leaves_home_skill_links_untouched` fails with `unexpected seeded crew copilot`.

`orbit init` probes the real `PATH` for provider CLIs (`agent_detect.rs`). `EnvGuard` isolates HOME, cwd, and the managed registry root — **but not `PATH`** — so this test's result depends on which agent CLIs the developer happens to have installed. It passes on CI and fails on my machine because the Copilot CLI is present.

The fix is an `EnvGuard::path()` builder alongside the existing `home()` / `cwd()`, but choosing what `PATH` a test should see is a change to shared test infrastructure with a real design choice in it, not a local repair. Flagging rather than deciding it unattended.

## Suggested follow-up (not included)

The class of bug here is "code only one platform compiles." A compile-only step would catch all of it for a fraction of a full macOS test run:

```yaml
- name: Compile all test targets
  run: cargo test --workspace --locked --no-run
```

Left out because macOS runner minutes are a cost decision, not mine to make.

## Verification

- `make ci-fast` — pass
- `make ci-lint` — pass
- `cargo test -p orbit-engine` — 412 pass, 0 fail (was: did not compile)
- `cargo test --workspace` — compiles everywhere; the one remaining failure is the `PATH`-dependent init test above

## Notes

No `ORB-` task ID — filed directly at Daniel's instruction while the box is down, which also overrides the repo rule against committing before task approval.

### DIFF
--- a/crates/orbit-cli/tests/mcp_roundtrip.rs
+++ b/crates/orbit-cli/tests/mcp_roundtrip.rs
@@ -1056,6 +1056,10 @@ fn generated_forced_command(workspace: &McpWorkspace) -> GeneratedForcedCommand 
 /// and the destination's callers row remains the grant ceiling. This executes
 /// the exact argv emitted into `authorized_keys`, through the real MCP stdio
 /// transport, for both sides of that intersection.
+// SSH MCP Tier 2 acceptance is a Linux-only deployment: the server refuses to
+// serve without the generated isolated-account setup, so this asserts on a
+// capability that does not exist on other platforms.
+#[cfg(target_os = "linux")]
 #[test]
 fn the_generated_forced_command_obeys_the_matched_rows_operator_ceiling() {
     let workspace = McpWorkspace::init();
@@ -1196,6 +1200,10 @@ ssh_key_fingerprint = "{CALLER_KEY_FINGERPRINT}"
 /// [ORB-11057] The public flag name is not proof that a destination generated
 /// the argv. Without the destination-issued value the old forged invocation is
 /// rejected before a caller row can be selected.
+// SSH MCP Tier 2 acceptance is a Linux-only deployment: the server refuses to
+// serve without the generated isolated-account setup, so this asserts on a
+// capability that does not exist on other platforms.
+#[cfg(target_os = "linux")]
 #[test]
 fn caller_controlled_accept_ssh_flags_cannot_select_a_caller_row() {
     let workspace = McpWorkspace::init();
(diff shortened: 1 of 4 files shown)


danieljhkim:
Correction to this PR's Verification section, for the record.

It said: *"`cargo test --workspace` — compiles everywhere; the one remaining failure is the `PATH`-dependent init test above."*

That was wrong. `cargo test` halts after a failing target, so the run I based it on never reached anything past `orbit-cli --bin orbit`. Once those passed, two more sets of failures appeared:

- **`mcp_roundtrip`** — 5 SSH MCP Tier 2 acceptance tests, ungated but exercising a Linux-only deployment (`Tier 2 requires the generated isolated-account SSH deployment on Linux`). Fixed in the follow-up commit on this branch by gating the tests; the shared `assert_command_succeeded` helper correctly stays ungated for its 9 non-SSH callers.
- **`workspace_sync`** — the same unresolved-vs-resolved fixture path issue as `shared_root`. Fixed in #1254.

The accurate statement at the time would have been: compiles everywhere, `orbit-engine` green at 412, remaining failures unknown because the run stopped early. I extrapolated from a truncated run instead of saying that.

PR #1666, opened 2026-09-08
fix(ci): isolate compiler cache namespace fixtures [ORB-11767]

danieljhkim:
Compiler-cache unit fixtures changed HOME and selected a fake sccache, then launched the live namespace suite with only PATH restored. On namespace-capable hosts this left the live compile without its real Rust toolchain.

Run the live suite as a separate process in the shared CI guardrails, immediately after the unit suite, so it inherits the caller environment. Reject leaked fixtures before capability skips, with regression coverage that fails against the previous namespace entry point.

Validation on macOS:
- Current head 3a9a89ccce7739d8fddd52a9121fc0d5d770b635: Apple Bash syntax checks and compiler-cache unit suite passed. Live Linux namespace compile explicitly skipped because bwrap is unavailable.
- The regression failed against the original namespace script as expected.
- Previous head e7eb3af546d128ae83c7235146bb8f8e6cb5f4ab passed make ci-fast and full workspace/all-target make ci-lint (16.51 seconds). Rebased onto integration refactor 613d4b56f; current head also passed full workspace/all-target make ci-lint. Native ci-fast has an independently tracked Bash 3 compatibility failure (ORB-11775); it is not claimed passing on this head.

Task: ORB-11767.

### DIFF
--- a/scripts/ci-guardrails.sh
+++ b/scripts/ci-guardrails.sh
@@ -69,3 +69,4 @@ fi
 "$repo_root/scripts/smoke-plugin-install.sh"
 "$repo_root/scripts/test-build-budget.sh"
 "$repo_root/scripts/test-compiler-cache.sh"
+"$repo_root/scripts/test-compiler-cache-namespaces.sh"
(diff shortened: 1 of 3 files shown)


danieljhkim:
@copilot resolve the merge conflicts in this pull request

